PR #7, opened 2020-08-19
Swich to ashcon v1 to solve issues

builder-247:
v2 currently breaks username loading as it returns 404 for every request


WarpWing:
Thanks for the assistance builder :D

PR #24, opened 2020-08-28
Update resources.ejs

Tydo690:
Imo a much better looking function considering the returns it makes.

Hope you like the function :(


apollyon600:
Looks good to me.

metalcupcake5:
Have you tested it? Just making sure it doesn't throw errors,

Tydo690:
> Have you tested it? Just making sure it doesn't throw errors,

Yes, i have tested it within my little set up. no errors

PR #25, opened 2020-08-28
make the method cleaner and also short

boblovespi:
I think this is the best of both worlds
its readable and it uses ternary


Shiiyu:
Looks good to me

WarpWing:
Sweet

Issue #19, opened 2020-08-28
Bug Report: duplicate talismans


Leobastian04:
You might already know this, but opening the missing Talisman menu Shows all Missing talismans, including downgrades of already owned Talisman, Like Pot affinity ring/talisman, Speed affinity talisman/ring, scarfs xxx(the Dungeon Xp boost one), Slayer defense talismen and rings, Red claw Ring/Talisman and some Other i might have missed, although i own the maxed version. Really Happy you included a "Missing Talisman" Feature though, thats one Thing i Always missed in skylea

metalcupcake5:
This is intentional, I merged this version into master. I just wanted to release it bcs it was taking a while. Expect a fix soon!

metalcupcake5:
Fixed #26

PR #26, opened 2020-08-29
Don't include lower teir talismans in missing list

metalcupcake5:
Lower talismans in a specific line that you already have don't appear in missing talismans. Solves #19


WarpWing:
Looks fine to me. Want me to merge?

PR #27, opened 2020-08-29
add emoji

boblovespi:


boblovespi:
> what is the point of the first emoji array

you're right, i'll get rid of it

Issue #28, opened 2020-08-31
Some enchantments like "Blast Protection" aren't fully colored


Untitled:
**EDIT**
There is also some OP special enchantments such as Growth L, would we add them to special enchantments list or don't consider these OP enchants? 

**Describe the bug**
Some enchantments, such as "Blast Protection VI" , which should be colored gold aren't colored completely.

**Profile/Location**
https://sky.shiiyu.moe/stats/Minikloon/Banana

**Expected behavior**
Fully Colored

**Screenshots**
![image](https://user-images.githubusercontent.com/33189830/91719282-894da900-eb84-11ea-8d4b-171ec15be8e3.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome

#29

WarpWing:
I would say they are special enchantments so we would add them in :D

PR #29, opened 2020-08-31
added some spec enchs

Untitled:
https://github.com/SkyCryptWebsite/SkyCrypt/issues/28

### DIFF
--- a/src/constants/misc.js
+++ b/src/constants/misc.js
@@ -99,7 +99,15 @@ module.exports = {
         'Feather Falling 10',
         'Feather Falling X',
         'Infinite Quiver 10',
-        'Infinite Quiver X'
+        'Infinite Quiver X',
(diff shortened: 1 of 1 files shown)


WarpWing:
Ah, this makes em fully colored in. Thank you for this commit :D Will merge shortly

metalcupcake5:
This won't fix the problem, it will detect prot 6 and go to color that, then move on.

WarpWing:
Hmmmm. My bad metal. I'll let you take it from here

apollyon600:
How did this become the problem if you forked it from the og website?
Unless you guys rewrote the code.

metalcupcake5:
It was a problem with Lea's site aswell, we just didn't fix it because nobody really uses blast prot, proj prot, and fire prot, much less the t6 versions of them

Issue #23, opened 2020-08-28
Possible issue with username


AnsonYeung:
**Describe the bug**
When a player username contains a underscore(I guess, I can give you the username for you to look into), the API will return `{"error":"TypeError: Cannot read property 'includes' of undefined"}`
However, the main page still shows the player data properly.

WarpWing:
I would like the username :D

AnsonYeung:
[Random Guy on Forum](https://sky.shiiyu.moe/api/v2/profile/ankles_)

WarpWing:
Mostly likely, it's is _ throwing us a error. Will look into though it probably won't be me. I need to look into other stuff. Mostly like it would be @boblovespi  or @metalcupcake5

WarpWing:
Looking over it now, it does seem like it's calling a method on a object that doesn't exist. Probably resolving something different that what's it's supposed to but that's just me <3

WarpWing:
API is currently on lockdown. I am closing this issue but it will be open in the future when API is back up

metalcupcake5:
API is back up, I'll try to look at this soon.

WarpWing:
@metalcupcake5  did you ever find a fix for this?

PR #65, opened 2020-09-17
Fix52

felix642:
Should fix the issue #23 when using the api or the website.

image
image
image


metalcupcake5:
Oops, I just fixed this. Sorry!

PR #68, opened 2020-09-18
+Few items from the api on the website, +Displays class level & dunge…

felix642:
I added a few items from the api
Players can see class level / dungeon level
Display stats about completed floors

image
image

Still need some work but that's a beggining


MartinNemi03:
Looks nice, but like you said, it still needs some work, but that's ok for now. I'll merge it later and rework some things in it.

AnsonYeung:
Class level stat is not accurate. [See the official data](https://api.hypixel.net/resources/skyblock/skills) under collections/DUNGEONEERING

MartinNemi03:
So, after a discussion within our team, we decided to use this only as an inspiration, since most of the backend code would need to be rewritten. I'll close this now, since we already opened a new pull request for this feature.

PR #73, opened 2020-09-22
Add first version of Dungeons with official API

MartinNemi03:
So, Hypixel finally released API support for Dungeons. 
Here we have the first version of Dungeons after their official support was added. 
Right now you can see: 
 - Classes and their levels
 - Catacombs with it's level and floor stats

Please review and give feedback.
Let's hope this doesn't break.


metalcupcake5 (approved):
Well its like 50% my code so ofcs i approve

src/lib.js
```diff
@@ -109,8 +109,8 @@ function getXpByLevel(level, runecrafting){
     return output;
 }
 
-function getLevelByXp(xp, runecrafting){
-    let xp_table = runecrafting ? constants.runecrafting_xp : constants.leveling_xp;
+function getLevelByXp(xp, type){
```
boblovespi, line 112: why is type an int?

boblovespi: also why type when there is another variable called type

PR #99, opened 2020-10-07
Adds client side live update to 'last-updated' and adds 'currently-upgrading-profile-upgrade'

doej1367:
adds an update cycle of 1 minute with a duration of up to 1 hour to the 'Last Updated: ' field.
grafik

NOTE: I could not test this for other timezones other than GMT+2.

(this also removes the code that tries to load the elevator music causing errors)

EDIT: now also adds the currently upgrading upgrade (has the same countdown mechanic)
grafik

here a picture of it when the countdown drops below a day (TimeDeo Profile):
grafik


metalcupcake5:
Could your code go into the javascript file?

FantasmicGalaxy:
I like the community upgrades countdown, and the rest of it looks good, so I'll approve it. But just for future reference, try to keep each pull request to only 1 feature. For example, the elevator.js stuff that you are removing has other parts that also need to be removed, but it'd be easier to do in just one PR instead of having the removal spread out over multiple.
Great job!

FantasmicGalaxy (approved):

metalcupcake5 (approved):

MartinNemi03 (approved):

nstringham:
This PR has some big merge conflicts due to es-lint prettier and typescript.

Should I do the necessary refactors or should we just close this PR?

PR #103, opened 2020-10-09
Manifest

nstringham:
adds a manifest and equivalent apple meta tags
- display:standalone and apple-mobile-web-app-capable allow mobile users to make better home screen shortcuts.
- manifest theme_color, the theme_color meta tag and apple-mobile-web-app-status-bar-style color the URL bar on mobile
- manifest icons and apple-touch-icon provide high-quality icons for the home screen and other purposes


FantasmicGalaxy:
I've got a few questions, but I ran a Lighthouse audit, and it seems to mostly check out. 
Screen Shot 2020-10-29 at 5 25 52 PM
Not sure about the splash screen, but it shouldn't need one. The icons should be fine because it looks like they are SVGs. I guess just that there doesn't seem to be a service worker?

nstringham:
>does not register a service worker...

and
>...does not respond with a 200 when offline

are outside the scope of this PR they refer to the fact that an installable PWA should work offline

>Manifest does not have a PNG icon of at least...

is caused by the fact that I only included webp and svg (not png) icons.
All of the browsers I tested supported webp app icons:
Chrome Windows: ✔
Chrome Android: ✔
firefox Windows: N/A
Firefox Android: ✔
Samsung Internet: ✔
iOS Safari: N/A

FantasmicGalaxy:
Ok, sounds good. I'll merge it now

PR #109, opened 2020-10-11
better scrollbar styling javascript

nstringham:
- javascript is now in resources.ejs instead of stats.js
  - scrollbar styling now works on index and stats
  - makes scrollbar styling load faster
- now runs directly after scripts are loaded instead of waiting for the load event
  - makes scrollbar styling load faster
- now makes its own elements rather than using document
  - makes scrollbar styling work when the viewport is taller than the content

### Before
image

### After
image


FantasmicGalaxy:
Can't actually see anything on mac, so someone else is gonna have to test it

FantasmicGalaxy (approved):

nstringham:
I don't have a Mac so haven't tested on Mac

FantasmicGalaxy:
Sorry, meant that I've tested it on Mac, and the styling doesn't look like it affects it.

nstringham:
@FantasmicGalaxy try emptying your browser cache

PR #111, opened 2020-10-13
Add favorites feature

metalcupcake5:
bad code go brrrrr


MartinNemi03 (changes requested):
Only thing, that bothers me, is this "Cookie 'favorite' will be soon rejected". 
If that's fixable, then I'm all for it, but if it's not, then merge it.

MartinNemi03 (approved):
Seems good to me.

PR #112, opened 2020-10-14
redo header styling

nstringham:
the header can be a little janky at times.

examples of the header being janky:
sky shiiyu moe_stats_Cookie_Wookie_7_Tomato
sky shiiyu moe_stats_Cookie_Wookie_7_Tomato (1)

To fix this I changed the header to use flexbox rather than putting position absolute on every element.

the header is now dynamic and responsive improving the experience for all users but especially mobile users.


FantasmicGalaxy:
Screen Shot 2020-10-29 at 8 24 06 PM
Screen Shot 2020-10-29 at 8 24 03 PM
Looks the same to me on the live site versus your changes, even when resizing the window?

nstringham:
@FantasmicGalaxy this is probably due to the old CSS being cached by your browser. 7c80241 should help.

PR #113, opened 2020-10-15
changed the pet candy count design to the in game one

doej1367:
NOTE: If there are no candies there is no counter.
grafik


boblovespi:
i think we dont want this right @Shiiyu

FantasmicGalaxy:
Doesn't quite look as clean imo, so I think that we'll just close this one for now

PR #121, opened 2020-10-20
Enable compression on responses from server

RDIL:
Small performance improvement.


LeaPhant:
Compression is already done in multiple places (reverse proxy and probably on cloudflare dns level too), this would only add a performance impact with no further reduction in size.

RDIL:
That's a fair point.  I'll run a profiler to see what is actually going on with the performance.

PR #147, opened 2020-11-19
Angler Armor Full Set

Pottedplant530:
This wont actually work its just a start yada yada ask metal

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -1440,7 +1440,9 @@ module.exports = {
             for(const stat in pet.stats)
                 output.pet_bonus[stat] = (output.pet_bonus[stat] || 0) + pet.stats[stat];
         }
-
+        // for angler armors full set ability being wrong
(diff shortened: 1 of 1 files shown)


Pottedplant530:
i forgot to say angler armors full set ability that gives 4scc was supposed to be removed but hypickle dumb

src/lib.js
```diff
@@ -1440,7 +1440,9 @@ module.exports = {
             for(const stat in pet.stats)
                 output.pet_bonus[stat] = (output.pet_bonus[stat] || 0) + pet.stats[stat];
         }
-
+        // for angler armors full set ability being wrong
+        if(items.armor.filter(a => getId(a).startsWith('Angler_Armor_')).length == 4)
+
```
Pottedplant530, line 1445: Obviously this wont do anything by itself this is is just a start to solve hypickles idoticy

PR #151, opened 2020-11-20
Pet abilities

boblovespi:


MartinNemi03 (changes requested):
Consider removing unnecessary things in header.ejs (line 8-35) and stats.ejs (line 2468-2554)

MartinNemi03 (approved):

PR #152, opened 2020-11-20
better Favorites

nstringham:
After many merging pains, I am proud to present a PR that should improve the multiple favorites system with zero merge conflicts.


Pottedplant530:
@metalcupcake5 this mans about to ruin your whole carrier

metalcupcake5:
hamsterspook

public/resources/js/stats.js
```diff
@@ -915,105 +915,45 @@ document.addEventListener('DOMContentLoaded', function(){
...
+        if(uuid == "0c0b857f415943248f772164bf76795c"){
+            favoriteNotification.setContent("No");
+        }else{
+            let cookieArray = parseFavorites(getCookie("favorite"));
+            if(cookieArray.includes(uuid)){
+                cookieArray.splice(cookieArray.indexOf(uuid), 1);
 
-                notification.show();
+                favoriteNotification.setContent("Removed favorite!");
+            }else{
```
metalcupcake5, line 944: I don't see a max amount of favorites here, is it set somewhere else?

metalcupcake5 (changes requested):
Besides my comment, everything else looks good.
I would prefer it to be a gold star when someone has been selected as a favorite, but that's just my preference

public/resources/js/stats.js
metalcupcake5: sigh it didn't select the correct parts whatever

MartinNemi03:
> 
> 
> Besides my comment, everything else looks good.
> I would prefer it to be a gold star when someone has been selected as a favorite, but that's just my preference

Yeah, that's your preference, we had a poll on Discord and this was ultimately desided.

public/resources/js/stats.js
nstringham: thanks, I forgot to add that it should be there now

Pottedplant530:
@metalcupcake5  are you ready to sign off on it?

MartinNemi03 (changes requested):
maybe rearrange things in getExtra, so it doesn't check the db for uuids of favorites on every page load (that's why it checks the db for favorites only on index). we had many problems with that before, so i dont think we want to reintroduce this issue.

nstringham:
@MartinNemi03 thanks for pointing that out it should be fixed in eee2796

metalcupcake5 (approved):

MartinNemi03 (approved):

Issue #155, opened 2020-11-21
extra scrollbar


nstringham:
**Describe the bug**
a scroll bar exists on the `#player_profile` element this is mast likely caused by the hover on fruit

**Profile/Location**
http://localhost:32464/stats/Cookie_Wookie_7/Tomato

**Screenshots**
![image](https://user-images.githubusercontent.com/44071655/99865936-3976f080-2b7b-11eb-9d34-6c8cf8d3ab75.png)

**Browsers**
- windows chrome latest
- ios safari latest
- android chrome latest
- windows firefox latest

PR #157, opened 2020-11-21
fix double scrollbars

nstringham:
this fixes two similar problems to each having to do with scrollbars showing up where they shouldn't

in the process of fixing these bugs, I also redid the basic stats section so it uses grid layout now


nstringham:
fixes #155

PR #166, opened 2020-11-23
Add prettier

RDIL:
- Adds the code formatter Prettier, which makes the code style look consistent
- Configures it to run prettier right before git makes a new commit, so its fully automatic


metalcupcake5 (commented):
Seems a bit much, not sure I approve of this. Some of the changes are good, but others seem to mess up different layouts we use.

PR #167, opened 2020-11-23
Add margin between enderchest pages

dukio:
Small css change to add 1rem margin between enderchest pages to improve readability.

Here's the result:

image

image


MartinNemi03:
Seems good, next step is doing proper pagination for that (which Lea wanted to do, but idk if she will xd), but this is still good for seperating it.

dukio:
> Seems good, next step is doing proper pagination for that (which Lea wanted to do, but idk if she will xd), but this is still good for seperating it.

Yeah I was thinking of that too.... this is just a quick way to improve enderchest. 

Actual pagination would be the top.
Honestly I might try to do that, but first I have to wait for the prettier/editorconfig fix.... even to make this commit I literally had to write the code in vscode, copy it, and paste/save the file using notepad to prevent vscode from fixing random spaces or mixed indentation :/

PR #178, opened 2020-12-10
NameMC link in profile

dukio:
image

image

I added it visible by default, I don't know if it should be hidden (and visible after clicking the ">" arrow)... I feel like there's a lot of links now; maybe we should hide some based on how often they are used? (do you have stats about this?)


metalcupcake5:
Why is your guild torturing bob

dukio:
> Why is your guild torturing bob

reference from Wynncraft. Also my former Guild Master is in a guild with the same name on wynncraft.

metalcupcake5 (approved):

MartinNemi03 (approved):

PR #200, opened 2021-01-02
Defer imgs

shadowtime2000:
Basic deferring of offscreen images like the ones found the navbar.

### DIFF
--- a/includes/footer.ejs
+++ b/includes/footer.ejs
@@ -7,6 +7,14 @@
 <script async src="https://arc.io/widget.min.js#oNMq8LVU"></script>
 
 <script>
+    var imgsToDefer = document.getElementsByClassName("defer-img");
+    function loadImg(img) {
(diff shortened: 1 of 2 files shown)


nstringham:
wouldn't it be better to use `loading="lazy"`

shadowtime2000:
@nstringham `loading="lazy"` loads when the image is a calculated distance away from the viewport, and since the navbar starts in view of the user, the images are just fetched normally.

nstringham:
the files deferred by this PR total 67 kb which adds about 0.007 seconds to page load time on 4G and they are cached so they will only be downloaded on the first page load.

I don't think the added complexity is worth the very small performance gains.

If these images are a problem I would recommend compressing the pack icons using a tool like Squoosh.

nstringham:
I recommend closing this PR in favor of #381

PR #203, opened 2021-01-04
Native Share

nstringham:
this PR adds a button to supported devices which when pressed will trigger the native share functionality.

image

Note: this feature should work on Windows, iOS, and Android but has only been fully tested on Windows due to HTTPS limitations

this PR also fixes a minor accessibility bug that affected several "buttons" in the `additional_player_stats` section.


metalcupcake5:
Is that a universal image for native share? How will users know to use it?

nstringham:
@metalcupcake5 that share icon is used by Material Design Android and Google so it's pretty universal. Additionally I use the Apple share icon if the user is on a Mac or iPhone

nstringham:
Tested and working on Android

PR #231, opened 2021-01-12
Fix Jerry talismans

dukio:
image


dukio:
Didn't notice that for some reason it doesn't include the missing talis into the total (it's still 64/64), will check

PR #241, opened 2021-01-18
Update max fairy soul count

userterminal:
Updating constant MAX__SOULS from the old 209 fairy souls to the new 220 (11 from Dwarven Mines)


RicardormDev (approved):
Seems to be ok.

PR #248, opened 2021-01-24
Titanium accessories

dukio:
They made the heads using transparent pixels, and skycrypt head renderer actually renders transparent pixels.... but in game they are black:
image
image

I'm not sure if this is something we should look into or just ignore

Anyway here's a preview with the real skins used in game:
image


metalcupcake5:
Oops I was going to do this but never had a chance, thanks!

metalcupcake5 (approved):

dukio:
> Oops I was going to do this but never had a chance, thanks!

no problem xD
Just got it in game and noticed it wasn't on skycrypt so... 🤷‍♂️

PR #255, opened 2021-01-29
elephant and monkey pet fortune abilities

dukio:
I'm not sure about the modifyStats() since those stats have not been implemented yet....

image
image

image
image


metalcupcake5 (approved):
I believe you did the functions correctly

metalcupcake5:
Which reminds me, we need to find a way to display those stats

PR #271, opened 2021-02-07
fix service worker cache

nstringham:
this PR doesn't do anything that users will notice but it allows us to make changes to the offline error message in the future


public/sw.js
```diff
@@ -1,22 +1,34 @@
...
-        caches.open(`skyCrypt-offline-v${VERSION}`).then((cache) => {
-            return cache.addAll([
-                '/resources/html/offline.html',
-                '/resources/img/bg.webp',
-                '/resources/img/logo_square.svg',
-            ]);
-        }),
+        caches.has(cacheName).then( async (exists) => {
+            if (!exists) {
+                console.log('reloading')
```
metalcupcake5, line 16: is this needed?

metalcupcake5 (approved):

public/sw.js
nstringham: I'm not sure what part you are asking about so I'll explain what they each do:

`caches.has(cacheName)` this code creates a promise that will return true if a cache with the same name already exists (aka false if we need to update the cache)

`.then( async (exists) => {` this code sets up an async context so that I can use `await` instead of `.then()` and it passes the value from `.has()` into this context as `exists`

`if (!exists) {` this makes it so that we only refetch the offline page when we need to instead of every time we change the SW

`console.log('reloading')` this is a debug line I forgot to remove see 55029d7e19df74e192bf58b1a379f99006fd27d7

metalcupcake5: oops i was trying to quote line 16 but ig it quoted everything

PR #284, opened 2021-02-14
Better flash

nstringham:
this PR redoes the way that an element is flashed when it is updated by selecting a different weapon so that the anime dependence is no longer needed for it and it uses CSS instead of js so it should run smoother.

eventually, anime can be removed if [scrollIntoViewOptions](https://caniuse.com/mdn-api_element_scrollintoview_scrollintoviewoptions) ever becomes widespread


nstringham:
This PR should not be merged untill after #286 is resolved

Issue #286, opened 2021-02-14
Sass version delay problem


nstringham:
**Describe the bug**
when the server is restarted and the index.css version is updated let's say from v10 to v11 and a browser requests v11 it will receive v10 but it will be called v11 then later when the server is restarted again to change it to v12 the browser will ask for v12 and will finally get v11 but it will be called v12

**Additional Info**
something (probably arc) is trying to minify our CSS I know this because it removes the scarce map comment at the end it also minifies our js btw

also this only happens on the main site not on localhost not on cupcake

**Expected behavior**
the CSS updates when the version number is changed

## CRITICAL WARNING ##
until this issue is resolved **DO NOT MERGE ANYTHING THAT MAKES CHANGES TO HTML AND CSS** until we get this fixed because it will cause problems. PRs that only change html or only change css are fine but PRs that change css will not take effect right away.

nstringham:
Known workaround: shut down express before doing git pull

PR #294, opened 2021-02-21
better iOS landscape

nstringham:
this PR redoes the way margin and padding are put on the right and left side of the stats page. this two main effects:
1. the overall layout of the stats page should be easier to maintain
2. iOS landscape looks way better

this PR also fixes several longstanding visual bugs with the info, themes, and packs boxes

## before:
IMG_0168
## after:
IMG_0167


nstringham:
Note: this PR should not be merged untill #286 is resolved

PR #295, opened 2021-02-22
fix catacombs skill bar

nstringham:
## before:
sky shiiyu moe_stats_Cookie_Wookie_7_Tomato

## after:
localhost_32464_stats_Cookie_Wookie_7_Tomato


nstringham:
Note: this PR does not require a server restart

Issue #275, opened 2021-02-12
npm i fails


MySixSenses:
**Bug**
Installing SkyCrypt locally fails at 
```sh
npm i
```
**Error Log**
https://mystb.in/WashingtonImplicationsWiki.basic

**Developer Tools**
 - Node.js version: v15.8.0
 - npm version: 7.5.1
 - OS: Raspberry Pi OS

**Additional context**
Everything I did before this
```sh
git clone https://github.com/SkyCryptWebsite/SkyCrypt.git && cd SkyCrypt
npm i
```

metalcupcake5:
Can you install the node-gyp deps from [here](https://github.com/nodejs/node-gyp)?

MySixSenses:
Then this happens
```

> SkyCrypt@1.0.0 postinstall
> node init

(node:3667) Warning: Accessing non-existent property 'collection_data' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
/home/pi/SkyCrypt/src/lib.js:3358
            const collectionData = constants.collection_data.filter(a => a.skyblockId == itemType)[0];
                                                             ^

TypeError: Cannot read property 'filter' of undefined
    at init (/home/pi/SkyCrypt/src/lib.js:3358:62)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
npm ERR! code 1
npm ERR! path /home/pi/SkyCrypt
npm ERR! command failed
npm ERR! command sh -c node init

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/pi/.npm/_logs/2021-02-13T02_18_14_240Z-debug.log
```

metalcupcake5:
hmmm interesting, i get that same error but it allows me to continue

metalcupcake5:
can you continue on with the installation as normal and see what happens?

MySixSenses:
Nothing happens, same error

metalcupcake5:
Running `npm start` results in the same error?

MySixSenses:
Yes

metalcupcake5:
Only things I can think of are either making sure you're on the latest version of master, and completely re-cloning the project.

PR #300, opened 2021-02-26
Fixed some crashing

MySixSenses:
This fixes the second problem in #275


nstringham:
What is the actual cause of the crashes?

MySixSenses:
```
TypeError: Cannot read property 'filter' of undefined
```
Probably missing properties or undefined variable

PR #308, opened 2021-02-28
Add dungeon grades

barrack-obama:
Here's a preview of what it looks like:
image

I hope this doesn't mess anything else up


MartinNemi03 (changes requested):
I wouldn't make it a big background image, maybe a small icon rather, since I don't think it's really to express the value of the grades. Also you're free to upload the grade into our resources folder, so we don't have to rely on some unknown external server to give us those images.

barrack-obama:
so it could be something like this maybe? 
image

and yes, I will be uploading the grades into the resources folder

metalcupcake5 (changes requested):
Same as what Martin said, please use the resources folder instead of an external site. It'll improve performance as well as making sure nothing breaks if the external site happens to go down.

nstringham:
These images really don't feel like the fit in with the SkyCrypt aesthetic to me. Maybe it would be better to just put it as a line under best run and use plan text instead of the images

barrack-obama:
image
So more like this?

barrack-obama:
bump

nstringham (changes requested):
barrack-obama#1

MartinNemi03 (approved):

metalcupcake5 (approved):
I'll merge when https://github.com/barrack-obama/SkyCrypt/pull/1 is merged

PR #329, opened 2021-03-12
revenant t12

dukio:
before
image

after
image

Also the total unique minions is hardcoded.... maybe it shouldn't, but for now it works I guess?


MartinNemi03:
Yeah, it'll be better when the max minion count gets automated.. _one day, hopefully_

PR #333, opened 2021-03-13
fix weight tooltip spaces

dukio:
Ok, I know this is kind of an useless PR... but seeing those uneven spaces was really triggering me :/

Before:
image

After:
image


metalcupcake5:
oh i didn't need to put those

PR #344, opened 2021-03-21
elephant monochrome

dukio:
image

skin will release tomorrow


metalcupcake5:
coming in clutch about 6m before skin releases

PR #345, opened 2021-03-21
pet exp progression

dukio:
Added the % progression (0 to 100) on pets tooltips:

image
image
image

Also I decided to let it go above 100% because some people (like me) can find it useful or nice to see.

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -2629,7 +2629,7 @@ module.exports = {
 
             lore.push(
                 '',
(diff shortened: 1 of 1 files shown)


metalcupcake5:
Does it differ based on rarity?

dukio:
> Does it differ based on rarity?

Yep, it uses the xpMaxLevel of the pet (5m for common, ..., 25m for legendary)

PR #346, opened 2021-03-22
Pets section update (show pets in inventories, show max 100)

dukio:
Pets section update:
- Includes the pets found in inventories (inventory, enderchest, etc...) in the pets section. It converts the real item object into a pet object like all the other pets (to make sure they all look the same).
- Shows max 100 pets (like in skyblock, when you can add max 100 pets in the pets menu). After 100 the pets are added in an expandable section "Show More Pets".
- Added a variable that controls the limit on the max number of pets to show (currently set to 250); this is to prevent pet spam.
- Rewrote how the pets are printed in `stats.ejs`.... just feels more organized now

image
image

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -525,6 +525,9 @@ async function getItems(base64, customTextures = false, packs, cacheOnly = false
         if(helper.hasPath(item, 'tag', 'ExtraAttributes', 'mined_crops'))
             item.extra.crop_counter = item.tag.ExtraAttributes.mined_crops;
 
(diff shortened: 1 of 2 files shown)


nstringham:
If someone has more than the maximum number of allowed pets how does it decide which ones to show

dukio:
> If someone has more than the maximum number of allowed pets how does it decide which ones to show

They are sorted by importance as always, the first 100 are shown directly, the rest is shown in the expandable section. After the 250 hard limit they are just removed.

PR #347, opened 2021-03-22
New slayer drops

dukio:
image

Adds the 2 new drops from Revenant 5:
- Shard of the Shredded
- Warden Heart

Changes default average drops from 0-100% mf to 0-200% mf (I would use the magicfind calculated by the stats but it's highly inaccurate, for me it says 33 while in game it's 149)

**This is a DRAFT:**
- Actual drop chances are unknown, for now I inserted numbers that seems to match what I actually dropped in game
- There's a "bug" in _Shard of the Shredded_, it adds an "s" at the end while it should be added after _Shard_

Does anyone know how to obtain the real drop chances?

### DIFF
--- a/public/resources/scss/index.scss
+++ b/public/resources/scss/index.scss
@@ -2650,7 +2650,7 @@ a.additional-player-stat:hover {
 
 .slayer-drops {
     margin-top: 10px;
-    height: 160px;
+    height: 200px;
     display: flex;
     flex-direction: column;
 }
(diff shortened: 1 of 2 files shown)


metalcupcake5:
Only way to obtain the drop chance is to have an admin state it

dukio:
> Only way to obtain the drop chance is to have an admin state it

...any idea how to do that? Anyone from SkyCrypt has a direct contact to an admin?

dukio:
We'll just remove the section soon

PR #351, opened 2021-03-24
Remove themes from the API page.

jaideng1:
Changing themes on the API page doesn't change the text color until a page reload, so it might just be better to not have a button to change the theme.


nstringham:
A better solution would probably be to just make the theme update instantaneously

metalcupcake5:
Closed in favor of #352

PR #354, opened 2021-03-25
remove suport for browsers without javascript

nstringham:
instead of supplying a bad experience to users with no javascript we just ask the user to enable it localhost_32464_(iPad)


metalcupcake5:
very nice design

PR #356, opened 2021-03-26
Rename certain mobs to reflect in-game names

MySixSenses:
It will make it easier to understand what the kills and deaths really are for, instead of having to guess at what it could be.
Sadan Statue => Terracotta
Obsidian Wither => Obsidian Defender

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -1988,6 +1988,34 @@ module.exports = {
 
             stat.entityName = entityName;
         }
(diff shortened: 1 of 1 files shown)


nstringham (approved):
Seams like a good idea.
Code looks neat and fits in with other already existing code.

metalcupcake5 (approved):

PR #364, opened 2021-04-01
Fix master catacombs weight

dukio:
This will fix the current issue with profiles who played master_catacombs to have weight NaN.

The weight calculator uses the xp accumulated for each dungeon to calculate its weight, the problem here is that master_catacombs it's not actually a new dungeon, but instead an extension of catacombs (the xp gained goes to catacombs, not master_catacombs).

This fix simply adds a check if the dungeon is a "master" and simply returns 0 for the weight (that will be added to the total `calculated.dungeonWeight`.

There's a chance Senither will will do it differently, and in that case we update the code; but until then this fixes the NaN issue.

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -859,7 +859,14 @@ function calcDungeonsClassLevelWithProgress(experience){
     return Math.min(level, 50);
 }
 
-function calcDungeonsWeight(type, level, experience){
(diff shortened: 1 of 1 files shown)


metalcupcake5 (approved):
commit message is nice and descriptive

PR #365, opened 2021-04-01
Handle weight error "NaN"

dukio:
This PR aims to handle any possible weight error "NaN".

It checks the weight variable and if it is falsy it gives it the value -1.
The total weight only sums the valid (>= 0) values, so this can never be NaN now.

While printing, it checks if the weight value is valid and if it's not valid (not >= 0) is just says "Error"

image

I know it's not a super elegant solution, but it's 3am and I can't think of anything better right now.

PS: I used -1 because 0 could be a valid value

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -2006,7 +2006,7 @@ module.exports = {
             });
 
         const random = Math.random() < 0.01;
-        
+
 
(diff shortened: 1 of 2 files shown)


nstringham:
dukio/SkyCrypt#2

PR #372, opened 2021-04-04
make plus on VIP+ gold

nstringham:
## before:
image
## after:
image


nstringham:
Also there may have been some edge cases where MVP+ and MVP++ we're displaying the incorrect plus color and if that bug did exist this PR fixes it

nstringham:
This bug also affects [LeaPhant/skyblock-stats](https://github.com/LeaPhant/skyblock-stats)

metalcupcake5 (approved):

PR #376, opened 2021-04-06
Removed x bit from 203 files

GodSaveTheDoge:
See the title, removed the x bit since it's not needed.
I'm too dumb to be able to count correctly, it's 203


nstringham:
corporate has asked us to identify the difference between these files

GitHub Files Changed tab: they are both empty

metalcupcake5:
What is the purpose of this?

GodSaveTheDoge:
>What is the purpose of this?
since they aren't meant to be executed, they shouldn't be executable. It's preferable to give only necessary permissions...

PR #378, opened 2021-04-06
Small Easter Egg

jaideng1:
Small chance that the player's model will start rotating. This serves pretty much no purpose, but it is fun to see.

### DIFF
--- a/views/stats.ejs
+++ b/views/stats.ejs
@@ -1104,6 +1104,22 @@ function jerriefy(rank) {
     </div>
     <div id="skin_display">
         <div id="player_model"></div>
+	<script>
+		function startRotate() {
+			let num = Math.floor(Math.random() * 1000);
+			let num2 = Math.floor(Math.random() * 1000);
+			if (num != num2) return;
+			
+			var deg_ = 0;
+			setInterval(() => {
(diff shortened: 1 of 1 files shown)


nstringham:
I have a few reservations for this
1. people will probably just think its a bug
2. is it really worth the performance overhead for something that only happens one in a million times
3. shouldn't easter eggs be unlocked by doing something not just getting lucky

nstringham (changes requested):
please don't inline non-critical js

nstringham (changes requested):
please animate css with css animations not with javascript setInterval

metalcupcake5:
>     3. shouldn't easter eggs be unlocked by doing something not just getting lucky

I've already added an easter egg that is randomized, so I dont have anything against that part.
I agree that it is unnecessary and a possible waste of a user's resources though

nstringham:
>I've already added an easter egg that is randomized

@metalcupcake5 Really what?

jaideng1:
> > I've already added an easter egg that is randomized
> 
> @metalcupcake5 Really? what?

I think it's the "snow monke" thing

jaideng1:
I'm going to close this, due to it being unnecessary

PR #379, opened 2021-04-06
Light themes and color refactor

nstringham:
This PR adds a light theme and performs several refactors to the color system

things this PR does:
- themes
  - restructure the themes constant
  - add stripes as an option for bars
  - add option for light themes
    - make the UI look good in light mode
    - use CSS variables for shades of grey to accommodate light mode
    - replace icons with SVG to allow changing colors
  - store logo color instead of image
  - remove "line" color
  - add light default theme
  - add candy cane theme
- replace 4 systems of color with 2 (kind of 3)
  - use dark color for rank
  - use dark/light color for stats depending on theme
  - use dark/light color for lore depending on theme
  - use dark colors for item rarity background
  - shift greyscale Minecraft colors instead of inverting extreams
- add more screenshots for manifest
- remove fallback for browsers with no webp support

fix #367
fix #247

### Candy Cane Theme:
localhost_32464_stats_8a3fa60d87aa4240bcdc624b90632529 (1)

### themes menu with new themes:
image

### Before:
sky shiiyu moe_

### After:
localhost_32464_

### After (light):
localhost_32464_ (1)

### Before:
sky shiiyu moe_stats_iijei

### After:
localhost_32464_stats_cb7608d414134bb3812bc0566664c771

### After (light):
localhost_32464_stats_cb7608d414134bb3812bc0566664c771 (1)

### Before:
sky shiiyu moe_stats_iijei (1)

### After:
localhost_32464_stats_cb7608d414134bb3812bc0566664c771 (2)

### After (light):
localhost_32464_stats_cb7608d414134bb3812bc0566664c771 (3)

### Before:
sky shiiyu moe_stats_iijei (3)

### After:
localhost_32464_stats_cb7608d414134bb3812bc0566664c771 (5)

### After (light):
localhost_32464_stats_cb7608d414134bb3812bc0566664c771 (4)

### Before:
sky shiiyu moe_stats_iiJei_Pear(iPhone 6_7_8 Plus) (2)

### After:
localhost_32464_stats_iiJei_Pear(iPhone 6_7_8 Plus) (1)

### After (light):
localhost_32464_stats_iiJei_Pear(iPhone 6_7_8 Plus)

### Before:
IMG_0202

### After:
IMG_0203

### After (light):
IMG_0204

PR #381, opened 2021-04-08
lazy load pack icons and theme logos

nstringham:
make image icons in the packs and themes selectors not load until the user opens the selector this will result in a very small performance gain

### DIFF
--- a/includes/header.ejs
+++ b/includes/header.ejs
@@ -64,7 +64,7 @@
             if(theme.hidden) continue; %>
         <label class="list-item">
             <% let icon = theme.logo.replace("../", "/resources/"); %>
-            <img class="icon" src="<%= icon %>" alt="">
+            <img class="icon" src="<%= icon %>" alt="" loading="lazy">
             <span class="name"><%= theme.name %></span>
             <div class="author">by <span><%= theme.author %></span></div>
             <input type="radio" name="theme" value="<%= theme_id %>">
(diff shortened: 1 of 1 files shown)


MartinNemi03:
Could we potentially lazy load item and head textures?

shadowtime2000:
@nstringham I don't think adding `loading="lazy"` would work, its not really a magic attribute. It only starts lazy loading if the image is off screen in the sense of being up or down and afaik doesn't really consider whether if it is hidden or being hidden behind something. It will just start loading and won't make a difference. 

Again, idk for sure.

nstringham:
> @nstringham I don't think adding `loading="lazy"` would work, its not really a magic attribute. It only starts lazy loading if the image is off screen in the sense of being up or down and afaik doesn't really consider whether if it is hidden or being hidden behind something. It will just start loading and won't make a difference.
> 
> Again, idk for sure.

I tested in chrome it works

do you really think I would submit a PR without testing?

shadowtime2000:
@nstringham 

I mean like I barely know you and haven't worked with you so how should I know? 

My message wasn't really sent with any rudeness considering the moments in which I pointed out it was just afaik and I didn't know for sure. 

On a side note: How exactly did you test a performance increase like this?

nstringham:
> Could we potentially lazy-load item and head textures?

this method only works with the `` tag and itesm are using background-image

I do however plan to refactor the way items and heads work once I am done with the themes refactor (#379) and I will definitely include some sort of lazy loading.

nstringham:
> @nstringham
> 
> I mean like I barely know you and haven't worked with you so how should I know?
> 
> My message wasn't really sent with any rudeness considering the moments in which I pointed out it was just afaik and I didn't know for sure.
> 
> On a side note: How exactly did you test a performance increase like this?

sorry I guess I was kinda aggressive
for this I used the network tab in chrome to observe that the images are not loaded until the selector is opened

shadowtime2000:
@nstringham Can you share a screenshot of the performance tab? I am interested in seeing the exact performance gains

nstringham:
@shadowtime2000 currently the Chrome performance tab crashes on SkyCrypt I used the network tab to validate that the images are being deferred

shadowtime2000:
@nstringham Oh sorry, I meant the network tab.

nstringham:
> Can you share a screenshot of the ~~performance~~ network tab? I am interested in seeing the exact performance gains

sure.

image
image
image
image

as you can see the performance gains are very small because the images are very small

you can also clone the repo if you want to test things out for yourself

MartinNemi03 (approved):

MartinNemi03:
I'll go ahead and merge it so it's on for the next server restart.

PR #395, opened 2021-04-13
Fix Technoblade PIG+++ color

### LINKED ISSUE #391: Number 1 priority should be this
**Describe the bug**
Technoblade's Pig rank doesn't show the correct colour of +'s
Its meant to be light blue

**Profile/Location**
https://sky.shiiyu.moe/stats/Technoblade/Blueberry

**Expected behavior**
Correct colour +'s

**Screenshots**
image

 - OS: Windows
 - Browser: Chrome
 - Version: 10

dukio:
fixes #391 

It was broken in PR #372 

While the PR code is correct, it seems like Hypixel ignores `"rankPlusColor": "DARK_PURPLE"` for PIG+++ rank, forcing it to be colour aqua. The old PR moved code that forced the +++ to be aqua into a constants file, and the color now gets overwritten by the `rankPlusColor` value found in the API.

This PR adds a check that ignores the `rankPlusColor` value from the API if the rank is PIG+++. I know it's not a super elegant solution but it's the only rank with this issue.... and only 1 player has it.... so.... 🤷‍♂️

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -500,7 +500,7 @@ module.exports = {
             result = "S";
         else
             result = "S+";
-        
+
         return result;
     },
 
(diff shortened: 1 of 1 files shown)


dukio:
use the other pr

PR #405, opened 2021-04-16
added rat pet

boblovespi:


boblovespi:
need to wait on hypixel api before testing this

dukio:
There's also need to be something done here to make it say Morph correctly:
https://github.com/SkyCryptWebsite/SkyCrypt/blob/31a3616178091d6e67ef0deb4bf3f2834e212e78/src/lib.js#L2555

image

dukio (changes requested):
isMorph and isMount should be const instead of let, otherwise all good

boblovespi:
image
note: health and crit damage have been swapped around

metalcupcake5 (approved):
hi yes it good

PR #406, opened 2021-04-17
Rarity upgrade corner

dukio:
Suggestion:
image

Preview in all the various sections where this applies:
image
image
image
image
image
image
image
image
image
image


public/resources/scss/index.scss
```diff
@@ -3956,3 +3956,27 @@ body {
...
+
+.piece.piece-rarity-upgrade-1 {
+    border: 0 dashed;
+    position: relative;
+    &::before {
+        content: '';
+        position: absolute;
+        top: 0;
+        right: 0;
+        width: 1rem;
```
nstringham, line 3968: why is it sized in rems?

nstringham:
I still think this is pointless because the item has an intrinsic rarity and you can just look as the lore if you really care but clearly a lot of people want this so ok

public/resources/scss/index.scss
dukio: I always use rems, it helps keeping a consistent styling between elements... here it doesn't really makes sense because everything else is in px but.... 🤷

nstringham:
this PR will have merge conflicts with #379 please @ me on discord and I'll resolve it

MartinNemi03 (approved):
Looks good to me. I'll request reviews from other devs, but from my perspective, it's a go from me.

nstringham:
@dukio dukio#3

metalcupcake5 (approved):

PR #420, opened 2021-05-01
Updating catacombs journals

dukio:
**Draft**: I'm currently trying to get them all, quest log doesn't say there's any more after these but I want to finish the last book to be sure about it.


dukio:
Confirmed total new journals is 200

PR #421, opened 2021-05-02
Master skulls

dukio:
Master skulls new talisman from master mode catacombs.


metalcupcake5:
completely forgot about this and committed it already :(

PR #424, opened 2021-05-04
Separate talisman upgrades into upgradeable and non-upgradeable 

metalcupcake5:
yea


MartinNemi03 (approved):
Let's hope Metal didn't do an accidental new-liner here..

PR #452, opened 2021-05-17
Storage!

dukio:
- Support for Storage
- Uses the correct backpack icons
- getItems() function now always return the skin instead of the resourcepack texture in case the item has a custom in-game skin (it was buggy with enderpacks for some reason)
- charged the position of the vault, I think it's more important than the other things on it's right (accessories are already above, and pots, fish bag, quiver are ....not so important)
- optimized inventory header section spaces on desktop to make space for Storage

image
image

### DIFF
--- a/views/stats.ejs
+++ b/views/stats.ejs
@@ -1862,54 +1862,53 @@ const getRarityUpgradeClass = item => {
                                         ></div>
                                         <div class="inventory-tab-name"><abbr title="Inventory">Inv</abbr></div>
                                     </button>
-                                    <%
 
-                                    if(items.storage.length > 0){ %>
+                                    <% if(items.storage.length > 0){ %>
                                         <button class="inventory-tab" data-inventory-type="storage">
                                             <div class="inventory-tab-icon item-icon icon-54_0"></div>
                                             <div class="inventory-tab-name">Storage</div>
                                         </button>
-                                    <% }
+                                    <% } %>
 
-                                    if(items.enderchest.length > 0){ %>
+                                    <% if(items.enderchest.length > 0){ %>
                                         <button class="inventory-tab" data-inventory-type="enderchest">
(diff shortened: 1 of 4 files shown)


views/stats.ejs
```diff
@@ -1852,51 +1852,63 @@ const getRarityUpgradeClass = item => {
...
                                             <div class="inventory-tab-icon item-icon custom-icon" style="background-image: url(/head/4cb3acdc11ca747bf710e59f4c8e9b3d949fdd364c6869831ca878f0763d1787)"></div>
-                                            <div class="inventory-tab-name">Quiver</div>
-                                        </button>
-                                    <% }
-
-                                    if(items.personal_vault.length > 0){ %>
-                                        <button class="inventory-tab" data-inventory-type="personal_vault">
-                                            <div class="inventory-tab-icon item-icon custom-icon" style="background-image: url(/head/f7aadff9ddc546fdcec6ed5919cc39dfa8d0c07ff4bc613a19f2e6d7f2593)"></div>
-                                            <div class="inventory-tab-name">Vault</div>
+                                            <div class="inventory-tab-name"><abbr title="Quiver">Quiver</abbr></div>
```
nstringham, line 1911: Quiver isn't an abbreviation for Quiver because it's the same word

nstringham:
on a profile that has less than 18 backpacks, only the backpacks are shown but it should probably show an empty square for the others.
image
screenshot taken on [Cookie_Wookie_7/Papaya](http://localhost:32464/stats/Cookie_Wookie_7/Papaya#Inventory)

dukio:
> on a profile that has less than 18 backpacks, only the backpacks are shown but it should probably show an empty square for the others.
> image
> screenshot taken on [Cookie_Wookie_7/Papaya](http://localhost:32464/stats/Cookie_Wookie_7/Papaya#Inventory)

Fixed with latest commit
image
image

MartinNemi03 (approved):

PR #476, opened 2021-06-01
Add mythic enderman pet

jxxe:
Added the enderman pet's mythic ability which grants 0.4% more combat xp from endermen per level. "xp" and "endermen" are not capitalized in the ability in game.

### DIFF
--- a/src/constants/petStats.js
+++ b/src/constants/petStats.js
@@ -876,6 +876,14 @@ class Enderman extends Pet {
 			desc: [`§7Increases your odds to find a special Zealot by §a${round(this.level * mult, 1)}%.`]
 		};
 	}
+
+	get fourth() {
+		let mult = 0.4;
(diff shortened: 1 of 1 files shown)


metalcupcake5 (changes requested):
There needs to be a check for mythic in the above `get abilities()` method in order for the fourth ability to be inserted

jxxe:
> There needs to be a check for mythic in the above `get abilities()` method in order for the fourth ability to be inserted

Done

nstringham:
@jxxe what profile can I use to test this

jxxe:
> @jxxe what profile can I use to test this?

Refraction has a mythic enderman pet

nstringham (approved):
looks good
image
image

metalcupcake5 (approved):

PR #482, opened 2021-06-05
remove slayer drops

nstringham:
This PR does 3 things:
- remove average drops
- refactor slayer styles (using the grid for kills)
- remove unused constants from `stats.ejs`
## Before:
image
## After:
image

### DIFF
--- a/public/resources/scss/index.scss
+++ b/public/resources/scss/index.scss
@@ -1675,8 +1675,7 @@ a.additional-player-stat:hover {
     color: var(--gold-hex) !important;
 }
 
-.stat-name.golden-text,
-.slayer-section-header.golden-text {
+.stat-name.golden-text {
     color: var(--maxed-hex) !important;
 }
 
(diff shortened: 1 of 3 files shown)


dukio (approved):

MartinNemi03 (approved):

FantasmicGalaxy:
Looks a little funky with the stuff being aligned towards the left, but not sure how you could make that better. Maybe only 3 columns so that the second row doesn't feel as empty?

FantasmicGalaxy:
Or maybe remove the total and center the t5 for revs?

dukio:
> Looks a little funky with the stuff being aligned towards the left, but not sure how you could make that better. Maybe only 3 columns so that the second row doesn't feel as empty?

My proposal was this but it was rejected
image

FantasmicGalaxy:
Huh. I like that better but it does still look kinda empty

MartinNemi03:
I agree that I like dukio's proposal. At least the centering of the elements.

nstringham:
I think centering looks really weird when it has an even number in the top row and an odd in the button

dukio:
so... what do we do?

MartinNemi03:
> so... what do we do?

I suggest merging it into master and if any further changes are approved, we can always open a new PR.

FantasmicGalaxy (approved):

dukio:
> > so... what do we do?
> 
> I suggest merging it into master and if any further changes are approved, we can always open a new PR.

Approve 👍

PR #488, opened 2021-06-05
voidling minion

dukio:


dukio:
Ok so in game it shows 12 tiers:
image

But there's no way to upgrad it past T11.

Revenant on the other hand can be upgraded to T12, but it's currently showing:
image

I think admins copy/pasted but edited the wrong one.... so yeah :/

PR #491, opened 2021-06-10
update api video

nstringham:
- video is now made with vanilla because that is something everyone can understand
- video is now 1080p+ instead of 720p
- added script to help in future
- use mp4 for a link to ensure it works for everyone including Safari users


nstringham:
it looks like GitHub can't preview video in files changed so here is the new video

https://user-images.githubusercontent.com/44071655/121521299-e7d06f00-c9c1-11eb-956c-d3056d03f45b.mp4

MartinNemi03:
LGTM!

PR #492, opened 2021-06-11
Cleaned up CodeQL and Updated CI Process

WarpWing:
A small clean update to the workflow system. Since there isn't much to be done regarding CodeQL, I simply just deleted some of the comments for visibility. I also cleaned up the SkyCrypt CI file a bit. It should now show the npm ci and npm build phases instead of just plain hiding it.

PR #494, opened 2021-06-12
es-lint and Prettier take 2 part 2

nstringham:
merge #493 and #492 before merging this PR

this PR adds CI to check whether a PR passes prettier and eslint.

note this PR does not include any changes to code files so it will fail.


nstringham:
This PR is waiting on #502

PR #495, opened 2021-06-12
eslint part 1

nstringham:
merge #493 before this PR

this PR makes a few changes to some code files to start fixing all the problems pointed out by eslint


nstringham:
the problem that CodeQL found was not caused by this PR

PR #513, opened 2021-06-13
reduce inline scripts

nstringham:
this PR moves some javascript from inline to seprate files wich decrease the page size. it also changes when some scripts exicute.


nstringham:
oh it also removes the warning that shows up on IE

PR #520, opened 2021-06-15
more realistic Enchant glint

nstringham:
This PR redoes the way the glint is applied to enchanted items. The new system doesn't use any JS instead it uses an SVG filter applied with CSS.

This performance improvement allows the enchanted glint to be applied to animated textures.

This PR also makes the enchant glow look like it does in-game except for on the SkyLea theme which uses a glint in the style of the old system. In the future, an option can be added to the settings menu to allow users to choose their preferred glint style.

This PR also removes the glint from heads to match in-game.

This system also makes animated glints possible but that is outside the scope of this PR.

Note: This PR does not fix potion enchantment glints that will come in a future PR.

Before this PR:
image

Not Enchanted:
image

After this PR:
image

After this PR (Lea Theme):
image


dukio:
I still think heads should also be enchanted :/

MartinNemi03 (approved):

PR #524, opened 2021-06-18
add cacheonly mode

nstringham:
this PR forces cache only mode on when the hypixel status API says that the status of `Public API` is `under_maintenance`

it also puts a variable called `cacheOnly` on extra to allow future commits to indicate to the user that the site is in cache only mode


MartinNemi03 (approved):
LGTM!

PR #530, opened 2021-06-19
wardrobe refactor

nstringham:
- ~collapsed wardrobe sets now show chestplate instead of helmet~
- wardrobe sets don't collapse on devices without hover
- empty wardrobe set slots now show a placeholder

before: image

after: image


dukio:
If the wardrobe slot has no chestplate, then it should show the first item in the set (helmet, leggings or boots).

Also why switch to chestplates? The head is most often the characteristic piece that identifies a set, otherwise it's just the armor color/texture.

nstringham:
> Also why switch to chestplates? The head is most often the characteristic piece that identifies a set, otherwise it's just the armor color/texture.

On many sets the helmet is something unique but the chest plate matches the rest of the armor so the chest plate is a better representation of the set of a hole. Additionally some helmets look like helmets but others are shaped like heads unlike the chest plates which have a unified shape

nstringham:
> If the wardrobe slot has no chestplate, then it should show the first item in the set (helmet, leggings or boots).

I thought about doing this but didn't for two reasons: It messes up the aesthetic of having all chestplate shapes things and It would have been a complicated solution that required more server-side code than I liked.

Neither of these things will be a problem if we decide to go back to showing the helmet as a preview.

dukio:
Thing is... it might look more aesthetically pleasing but it ruins a lot of usability.

Example, looking at this:
image

good luck trying to understand what the armor actually is... that yellow chestplate, is it sponge chestplate? is it an exotic shark scale? is it any other exotic colored yellow? or is it a superior dragon?
the black one... a tuxedo? an obsidian chestplate? a wither chestplate? a shadow assassin chestplate?

then you look at the helmets and in half a second you know exactly what everything is
image

nstringham:
reverting chestplate change because
- not enough people supported it
- in some texture pack a lot of chestplates look similar and helmets are generally more iconic
- it will make the system less complex allowing for the easy addition of other features such as showing the first item in a set

metalcupcake5 (approved):

dukio (approved):
lgtm

PR #532, opened 2021-06-19
grammer and accesablity

nstringham:
- removes "an" from "an API maintenance"
- change "user's" to "users'" because more than one user
- put only one `<br>` per line
- use `<strong>` instead of `<b>` for improved accesablity


metalcupcake5:
*grammar

PR #539, opened 2021-06-22
Client side Typescript

nstringham:
This PR is similar to #250 but with TypeScript

TypeScript is a superset of Javascript that helps developers to write better code. think of it like linting on steroids.

This PR only converts `/public/resources/js` to typescript. It does not affect `/src`.

Because this PR includes many small refactors it's possible that one of them includes a mistake. I tried my best to test everything but I might have missed something.

the config for TS is no implicit any and strict. This means that all variables must be typed.

this PR also adds a CI job to check for typescript and sass compile errors.

this PR also makes the `npm run dev` command automatically run sass watch and typescript watch


MartinNemi03 (approved):
LGTM!

PR #540, opened 2021-06-22
script type module

nstringham:
This PR should be merged after #539 

This PR switches client-side code to es-modules. This will allow for tools like rollup to be added in the future. It also makes it easier to write good code by making it clear what script depends on what other script.


nstringham:
marking as draft because of bug where common-defer.js is loaded twice because one doesn't have hash

nstringham:
> marking as draft because of bug where common-defer.js is loaded twice because one doesn't have hash

this is fixed in #554

nstringham:
closing in favor of #554

PR #542, opened 2021-06-22
use API to determine API status

nstringham:
this PR changes the way `forceCacheOnly` is determined

a request is made to `https://api.hypixel.net/skyblock/profiles` for the profile of hypixel the server owner

if a javascript exception is thrown `forceCacheOnly` is set to true

if the response code is between 200 and 299 it is set to false

if the response code is 429 it is set to false because this is a key throttle

in any other case, it is set to true


MartinNemi03 (approved):
Looks good to me, but idk why, I'm thinking about suggesting 3-5 minutes between requests. Maybe a minute is fine, just thought I would leave this thought there.

nstringham:
I picked a minute somewhat arbitrarily if people think a different time would be better I'm happy to change it

A shorter time would respond more quickly to an outages

A longer time would use less system resources

PR #554, opened 2021-06-26
rollup

nstringham:
use rollup for client-side scripts

here is what rollup does in this PR:
- compiles typescript to javascript
- hashes file names
- bundles node modules
- minifies javascript

the main advantage of rollup in this PR is that we can use skinview3d as an npm module instead of manually creating and update the bundle

this PR also includes the changes from #540 which switch client-side script to modules


nstringham:
I am not done with everything I want to do with rollup but This branch is in a good place to merge to I'm opening this PR and I'll make another with further changes

PR #578, opened 2021-07-04
change cape cache times

### LINKED ISSUE #574: Bad optifine cape cache timings
On the server, Optifine capes are cached for only 10 seconds
https://github.com/SkyCryptWebsite/SkyCrypt/blob/63c445c495754c28639f7ea79ead8d34cf3705e7/src/app.js#L347

but on the client, they are cached for 30 days
https://github.com/SkyCryptWebsite/SkyCrypt/blob/63c445c495754c28639f7ea79ead8d34cf3705e7/src/app.js#L372

10 seconds is too short and 30 days is to long

nstringham:
this PR changes the longest time that it could take for an optifine cape update from 0-30.5 days to 0-24 hours

it also does some other fixes and optimisations in the cape endpoint

fix #574


nstringham:
marking as a draft for now because these changes don't make sense if nginx caches for 12 hours

nstringham:
@WarpWing I can ignore the CodeQL rate limit failure right?
https://github.com/SkyCryptWebsite/SkyCrypt/pull/578/checks?check_run_id=2985679611

PR #587, opened 2021-07-06
Update themes.js

JSanchoG:
New Theme - Bloody Scorpion
bg
bg_blur

PR #590, opened 2021-07-07
New Theme - Bloody Scorpion

JSanchoG:
skyshiiyemoe_index
skyshiiyumoe_player


JSanchoG:
I've been trying to get a different screenshot, but this enchanting table from my skyblock island fits there **perfectly** (not many sceneries could match dark red theme like this, I made over 40 screenshots in different skyblock locations, but still couldn't make it any better, because they were too bright or there was just full black color in blurred section (but in my original screenshot there are some nice green leaves that matches everything));

The best screenshot from blazing fortress I could make (still doesnt match this good):
bg

How site looks with the new (worse) screenshots:
look-1
look-2

nstringham:
Related to #587

JSanchoG (commented):
yes.

nstringham (changes requested):
It apears the screenshots in this PR are not accurate. here is what it actually looks like on ce2ee247c552f4340ba52f97d9e1ab0835bcad34 Firefox 89
image
image

## There are some technical and some content problems with this PR

### Technical problems
these problems are a techinal error
- background type `linear-gradient` is not a valid type if you want a gradient you'll need to add it as an option [here](https://github.com/SkyCryptWebsite/SkyCrypt/blob/master/public/resources/ts/themes.ts#L60-L71)
- bg-blur.png is not required

### Content problems
these problems are not a errors but they make the theme not fit in well with the other skycrypt themes
- background is not a screenshot of a public skyblock island
- background does not use 16x texture pack
- `link` is a different color than all the other colors
- `hover` is darker than `link`
- `maxed` is to dark
- `maxed`, `gold`, and `maxedbar` are not yellow

nstringham:
@JSanchoG if you don't intend to fix the errors on [JSanchoG:patch-1](https://github.com/JSanchoG/SkyCrypt/tree/patch-1) you should close this PR

PR #614, opened 2021-07-15
Gem collection texture

### LINKED ISSUE #610: Add correct textures for new collections
Screenshot_20210714-221409~2

nstringham:
close #610

This PR replaces the temporary red wool texture with the texture from the ~~ROUGH JASPER GEMSTONE~~ FINE JASPER GEMSTONE

image

### DIFF
--- a/src/constants/collections.js
+++ b/src/constants/collections.js
@@ -291,7 +291,6 @@ module.exports = {
       damage: 14,
       maxTier: 11,
     },
-
     {
       type: "combat",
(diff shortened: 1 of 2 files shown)


nstringham:
marked as draft because it's not actually the same texure

dukio:
Can't check in game right now but doesn't the Gemstone collection uses Ruby Gem?

nstringham:
> Can't check in game right now but doesn't the Gemstone collection uses Ruby Gem?

looks like jasper to me
image

PR #652, opened 2021-07-21
MongoDB error fix `TypeError: cannot read property 'date'/'last_updated' of undefined`

dukio:
Basically the code was expecting a query with no results found to return `null`, while for me in localhost it returns `undefined` instead.

But why is it working with `null` on Nate PC and on the server? 🤔

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -119,7 +119,7 @@ module.exports = {
       }
     }
 
-    if (cacheOnly === false && (user === null || +new Date() - user.date > 7200 * 1000)) {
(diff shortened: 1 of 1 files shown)


nstringham:
Wouldn't it be better to use ` == undefined` so that we are checking for nullishness instead of truthiness

dukio:
> Wouldn't it be better to use ` == undefined` so that we are checking for nullishness instead of truthiness

Updated using ` == undefined` or `!= undefined`

nstringham (approved):
you could also do `== null` if you want to change less but this works

nstringham (approved):
I tested this on my PC and everything works

PR #658, opened 2021-07-22
Add alert for outdated browsers

nstringham:
the most common problem for SkyCrypt users that ask for help on discord is an outdated browser and there are probably more users that don't come to the discord and just give up on SkyCrypt. After this PR SkyCrypt will alert users using an old browser and link them to a resource to help them solve the problem.
- on desktop, it links to the official help page for updating that browser
- on iOS, it links to the help page for updating iOS
- on Android, it links to the play store listing for the browser they are using

if the user is using a browser compatible with SkyCrypt it does nothing.
if the user is using a browser other than iOS, Chrome, Edge, and Firefox it does nothing

in the future when we make a change that breaks compatibility with old browsers we can update this list

it will be easy to add new browsers to this list in the future such as Brave, Opera, and Samsung Internet

image
image
IMG_0242
Screenshot_20210721-234201


MartinNemi03 (approved):
LGTM! 
Good job on this one Nate, let's hope it reduces traffic in website-support :D

PR #690, opened 2021-08-02
hatching pets and small pets refactor

### LINKED ISSUE #689: Golden dragon skin + name when below level 100
When below level 100, Golden Dragon has:
- Skin: https://sky.shiiyu.moe/head/113bdf2d2b00605606826df76e211ea288aa050edc9d71cb09986c488ca0411c
- Name: Golden Dragon Egg

image

dukio:
So I wanted to make hatching golden dragon have "egg" in the name and a different skin and then I got carried away a little...

This PR adds/changes:
- Pets can now have a `hatching` property that includes `level` (up until they are at the hatching stage), `name` (different name while hatching) and `head` (different head texture while hatching). Golden Dragon is the first pet of this kind, I tried to make something that can be reused in the future 🤞 
- Moved "Morph" and "Mount" definitions in pets.js constants instead of hardcoded in lib.js, now pets in pets.js can have a new property `category` (feel free to suggest a better name, I didn't want to use _type_ as it's already being used for the pet type (dragon, pig, endermite, etc...)
- petStats now uses the pet.type instead of the pet.name to match pet => class (had to change this to make it work with different name while pet is hatching)... and since I was changing that I also sorted the names ascending to make it prettier 😛 
- At line 2988 of lib.js (before changes) there was an if() that didn't make any sense to me, since it was either looking for a pet.type in petStats or "???", which we know it's in there...

(Fixes #689 )


nstringham (approved):
LGTM

metalcupcake5 (approved):

MartinNemi03 (approved):

PR #699, opened 2021-08-03
Support for new stats (mining speed, all fortunes, pristine)

dukio:
Support for:
- Mining Speed
- Mining Fortune
- Farming Fortune
- Foraging Fortune
- Pristine

Changes:
- Updated Ability Damage symbol/icon

Preview:
image

...yes, it takes quite a lot of space. I would add a "show all" after Intelligence (so everything below that is hidden by default), thoughts?

I also updated the stats everywhere (I hope) and I noticed the stats could really use a refactor with a constants/stats.js that contains all the hardcoded things (prefix, suffix, symbol, shortname, longname, etc)


metalcupcake5 (approved):

metalcupcake5:
weed number

nstringham:
Looks good at first glance

I agree that it's taking up too much room but I'm not sure hiding some stats by default is the best solution

MartinNemi03:
Maybe we could hide stats that are empty (like Mining Speed and Pristine in dukio's screenshot)

PR #711, opened 2021-08-05
fix crit damage

dukio:
fixes crit damage missing bug


nstringham:
did I break that?

dukio:
> did I break that?

...yap, forgot to put the value for that field 🤷

nstringham:
Well good job Nate you did it again

PR #712, opened 2021-08-05
fix dungeon weight calculation error

dukio:
This PR:
- Adds a small red `!` next to weight if one of the 3 weights failed to calculate.
- Formatted the code that was printing the Weight stat and tooltip.... turns out there was an extra `</span>` there lol
- Fixes a bug where if a player never played a class, their entire dungeon weight will be Error.

The issue was caused by the player_classes looking like this:
image

I first tried to skip calculation for classes never played, but all the code after it and in frontend expects the resulting object to have all the classes so I ended up manually setting experience=0 for the never played classes.

Ref
image

### DIFF
--- a/public/resources/scss/index.scss
+++ b/public/resources/scss/index.scss
@@ -4178,3 +4178,7 @@ body {
   font-size: 0.8em;
   letter-spacing: 3px;
 }
+
+.stat-error {
+  color: var(--§4);
+}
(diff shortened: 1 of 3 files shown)


nstringham (approved):
Looks good didn't test

MartinNemi03 (approved):

PR #714, opened 2021-08-06
Gemstone stats

### LINKED ISSUE #698: Include gemstones added bonus in lore and player stats
Each gem type and tier gives a different stat bonus, would be pretty useful to display it!

dukio:
Spreadsheet where I'm gathering the values: https://docs.google.com/spreadsheets/d/1KAzrdT5aFaQMFDf2mHIKrPwPnLmIJebQu7C3Va9FR9s/edit?usp=sharing

So I wrote down all the values from the wiki only to realize too late they are wrong 😭 

In the next days I'll manually gather the correct values and update this, making this draft just so I don't forget

Fixes #698


dukio:
PR ready:
- Items with gemstone now display the given stat bonuses (gems already support multiple stats, even though that's not the case right now)
- I manually checked most of the gems and their bonus on items, but I couldn't get them all... for the ones missing it will display "MISSING VALUE"
- In lib.js I made what is possibly a **big change**, I moved the code that generates `item.extra` below the one that generates `item.type` and `item.rarity` since I needed the rarity to generate the extra.gems lore. From my tests everything seems to be working fine without any errors
- Moved stat symbols into their own ./constants/stats.js file (I will improve this in the future, but for now I did this to not duplicate the symbols object that was previously in petStats.js)

dukio:
Preview on a Gauntlet (which is the only case/item where showing gems stats is useless... but at least you can see the values are correct)
image

Preview when a gem stat value is missing
image

Preview of how it will look like most of the time
image

MartinNemi03 (approved):
LGTM!

PR #717, opened 2021-08-08
allow stats to be cached between users with different favorites

nstringham:
this fixes a thing @MartinNemi03 did


MartinNemi03 (approved):

MartinNemi03:
thanks, LGTM!

Issue #722, opened 2021-08-09
Add redis configuration support


Flou21:
**Is your suggestion related to a problem? Please describe in detail.**
A clear and concise description of what the problem is.
I tried to set up skycrypt, but I only have a remote redis instance to use.
Currently there is no option to configure the redis hostname.

**Describe your suggestion**
Add an option in the `credentials.json` file to add this configuration option

**Describe alternatives you've considered**
I forked this repository and implemented the change. I will open a pull request, but this would be my first real contribution to any open source project, so if I made some mistakes or my change does not make sense for you, feel free to decline the pr.

**Additional context**
When no `redisUrl` is set in the `credentials.json` file nothing should change, otherwise use the given config

PR #723, opened 2021-08-09
add redis configuration support #722

Flou21:
would solve issue #722 

please correct me if I made somthing wrong

### DIFF
--- a/src/apiv2.js
+++ b/src/apiv2.js
@@ -6,7 +6,7 @@ const constants = require("./constants");
 const credentials = require("./../credentials.json");
 
 const Redis = require("ioredis");
-const redisClient = new Redis(credentials.redisUrl ?? credentials.redisUrl);
+const redisClient = new Redis(credentials.redisUrl);
 
 function handleError(e, res) {
   console.error(e);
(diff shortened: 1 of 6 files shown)


src/apiv2.js
```diff
@@ -3,9 +3,10 @@ const lib = require("./lib");
 const cors = require("cors");
 const sanitize = require("mongo-sanitize");
 const constants = require("./constants");
+const credentials = require("./../credentials.json");
 
 const Redis = require("ioredis");
-const redisClient = new Redis();
+const redisClient = new Redis(credentials.redisUrl ?? credentials.redisUrl);
```
nstringham, line 9: `credentials.redisUrl ?? credentials.redisUrl` is redundant because `credentials.redisUrl` and `credentials.redisUrl` are the same thing.

see this MDN page if you're not familiar with how `??` works https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

nstringham (changes requested):

src/apiv2.js
Flou21: thanks for the quick reply and sorry for that stupid mistake

Flou21:
I close this because #755 also fixes my problem

PR #724, opened 2021-08-09
Lit and inventory refactor

nstringham:
this PR adds [Lit](https://lit.dev/) to SkyCrypt and uses it to rework the inventory.

adds a map from itemId to item is stored to make accessing items by Id easy

the grid of inventory items is now not processed on the server at all

the main inventory grid and the backpack preview in the lore now share code

more code is loaded asynchronously

code is easier to read


nstringham:
it's bean a while sense I worked on this so it might not be entirely done.

I think this is ready to merge but please test this toughly before merging

dukio:
Tested with multiple profiles and everything looks good.

I noticed you didn't use scoped styles, any reason for that? Should we not use that in the future?

nstringham:
> I noticed you didn't use scoped styles (inside the component .ts), any reason for that? Should we not use that in the future?

In the future scoped styles and shadow root would be great but converting an existing element to shatterroot can be difficult.

The main reason I didn't use shadow root Is because the items in the inventory share a lot of CSS with the items in the weapons/armor/pets sections. I had originally wanted to also convert the items in those other sections to use the `rich-item` element but I have decided That is outside the scope of this PR.

dukio:
tested, looks good :thumbsup:

dukio (approved):

MartinNemi03 (approved):

metalcupcake5 (approved):

PR #732, opened 2021-08-12
added the blocks_walked counter from the prehistoric egg

doej1367:
This code adds blocks_walked counter to the item lore of the prehistoric egg including the amount of steps needed for the next rarity level.
grafik
grafik


src/constants/misc.js
```diff
@@ -165,6 +165,9 @@ module.exports = {
   // Number of kills required for each level of expertise
   expertise_kills_ladder: [50, 100, 250, 500, 1000, 2500, 5500, 10000, 15000],
 
+  // Number of kills required for each level of expertise
```
nstringham, line 168: This comment is not accurate

src/lib.js
```diff
@@ -574,6 +574,14 @@ async function getItems(base64, customTextures = false, packs, cacheOnly = false
...
     }
 
+    if (helper.hasPath(item, "tag", "ExtraAttributes", "blocks_walked")) {
+      let { blocks_walked } = item.tag.ExtraAttributes;
+
+      if (blocks_walked > 0) {
+        item.extra.blocks_walked = blocks_walked;
+      }
+    }
+
```
nstringham, line 584: Why do we need this section can't we just use the value from the tag when we need it later on instead of storing redundant information in extra

src/lib.js
```diff
@@ -662,6 +670,26 @@ async function getItems(base64, customTextures = false, packs, cacheOnly = false
...
+            itemLore.push(`§8MAXED OUT!`);
+          } else {
+            let toNextLevel = 0;
+            for (const e of constants.prehistoric_egg_blocks_walked_ladder) {
+              if (blocks_walked < e) {
+                toNextLevel = e - blocks_walked;
+                break;
+              }
+            }
+            itemLore.push(`§8${toNextLevel} steps to tier up!`);
```
nstringham, line 688: "steps to tier up" sounds weird why don't we just put one line like "blocks walked 789/1000" and let people do the subtraction in their head

nstringham (changes requested):

src/lib.js
doej1367: Is "walk x blocks to tier up" better? I wanna keep the calculating to the computer.

src/lib.js
dukio: It's the same way expertise kills are done (and other things too iirc)... shouldn't be too much of an issue.

Normally item.extra has a "parsed" version of item.tag.ExtraAttributes, but we usually just store everything there, even if no parsing is required.

dukio (approved):
lgtm didn't test

src/lib.js
doej1367: > storing redundant information in extra
how is "blocks_walked" redundant info? It's basically the same procedure like for the expertise_kills. If I delete the lines you are reffering to the added lore does not show at all... Couls you clarify what you mean by "why do we need this secion". If you mean what I think you do then we don't need the corresponding lines of code for the other tags - like expertise_kills, rarity_upgrades, etc. - either. But then the nice and tidy seperation between the information extraction and the lore creation gets broken up.

src/constants/misc.js
doej1367: thx fixed

src/lib.js
nstringham: > how is "blocks_walked" redundant info

It's redundant because it is already stored in tag so it doesn't really need to be stored in extra

However if it is the same as `expertise_kills` then it should be fine we can refactor it later but we should stay consistent by keeping similar systems codded similarly

dukio: Yeah how he did it it's consistent with current code 👍 

And yes it could use a refactor.... together with pretty much half of lib.js 😛

src/lib.js
nstringham: The more I think about it the more I think saying the remaining blocks is a good idea but "to tier up" still sounds kinda weird to me but I'm not sure what would be better

nstringham (approved):
There are some code decisions I don't like but I agree that we should keep it the same as expertise kills so this PR is good but I want to refactor this later

Didn't test

MartinNemi03 (approved):

MartinNemi03:
Went ahead and resolved a merge conflict from changes in #714 in order to merge this PR.

PR #755, opened 2021-08-23
Thread and Database Abstraction and Optimisation

nstringham:
Changes:
- only run app.js on non-master threads
- only run scripts on master thread
- only make one MongoDB instance per thread
- only make one Redis instance per thread
- allow redis URL to be configured and update Docs

closes #722 
closes #723

PR #761, opened 2021-08-24
ECMAScript Modules

nstringham:
#755 and #764 should be merged before this

This PR changes the server-side javascript files from CommonJS modules to ECMAScript modules.

CommonJS is a standard used by Node because Node was invented before es6.

Now that es6 is a thing we have ECMAScript in Node.

ECMAScript modules have a number of advantages from CommonJS modules including top level await and named exports.

in addition to switching to ECMAScript modules this PR also:
- creates a small wrapper around `credentials.json`
- requires file extensions for imports
- removes `init.js` and moves initialization code to `master.js`

Another key advantage is that Typescript IDEs like VScode can now see values inside of constants

before:
image

after:
image


nstringham:
As far as I can tell none of the CodeQL problems are actually caused by this PR

MartinNemi03 (approved):

PR #765, opened 2021-08-26
less `getPath`, `hasPath`, and `setPath`

nstringham:
in this PR I refactor code so that it doesn't use the `getPath`, `hasPath`, and `setPath` helper function I replace these with modern vanilla js future like optional chaining and the `in` operator.

the only remaining uses of these functions are the lines where an array is being exploded into the function

I did not modify `api.js` because it is outdated anyway


metalcupcake5 (approved):

MartinNemi03 (approved):
LGTM!

PR #781, opened 2021-09-01
cachebusting stats icons

dukio:
Just adding `?v2` to all stats icons for cache busting

Didn't test... I can't test, I don't have the old images in cache


nstringham (approved):
Looks good didn't test

MartinNemi03 (approved):

Issue #204, opened 2021-01-04
tippy tool tip shows up in the wrong spot


nstringham:
**Describe the bug**
some of the tooltips are positioned incorrectly

**Profile/Location**
any

**Additional Info**
this bug also affects skylea

**Expected behavior**
tooltip points to the thing hovered over

**Screenshots**
![image](https://user-images.githubusercontent.com/44071655/103584827-147ed880-4eb0-11eb-9b80-4b131c769d9d.png)
![image](https://user-images.githubusercontent.com/44071655/103584826-134dab80-4eb0-11eb-99b9-e10222547c79.png)

**Browser:**
any

**Additional context**
this bug is caused by tippy trying to fit the tip within its parent that has overflow set

fixed by #161

metalcupcake5:
Fixed with #161

PR #784, opened 2021-09-01
Import Tippy 6 directly

nstringham:
This PR makes tippy an npm dependency instead of using the CDN. this allows us to tree shake and also means that the browser doesn't have to open an extra connection. it also means that dependabot can update tippy now.

This PR also updates from tippy 5 to tippy 6

### DIFF
--- a/public/resources/ts/common-defer.ts
+++ b/public/resources/ts/common-defer.ts
@@ -41,7 +41,7 @@ document.querySelectorAll<HTMLFormElement>(".lookup-player").forEach((form) => {
     } catch (error) {
       const errorTip = tippy(form.querySelector("input") as HTMLInputElement, {
         trigger: "manual",
-        content: error || "please enter a valid Minecraft username or UUID",
+        content: (error as string | undefined) ?? "please enter a valid Minecraft username or UUID",
       });
       errorTip.show();
       setTimeout(() => {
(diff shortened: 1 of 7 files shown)


public/resources/ts/common-defer.ts
```diff
@@ -143,9 +142,7 @@ function setCheckedTheme(theme: string) {
 
 setCheckedTheme(localStorage.getItem("currentTheme") ?? "default");
 
-tippy("*[data-tippy-content]", {
-  boundary: "window",
```
dukio, line 147: any clue why this was using `boundary: window` instead of the default `scrollParent`?

nstringham: I remember it was me that added it and it fixed a bug that we had been trying to fix for a long time but I can't remember the details

nstringham: here is where I did it https://github.com/SkyCryptWebsite/SkyCrypt/commit/eac7a231b62d5c9e9b080b2d444eb2160a8fe531

nstringham: it seams to be fixing #204

nstringham: after some testing this PR does not appear to reintroduce #204

MartinNemi03 (approved):

PR #785, opened 2021-09-02
fix: :lipstick: Prevent skill icons from clipping

bishoperic:
Before, skill icons were clipping the outside and it didn't look very good. The `transform: scale()` is optional, it just was there before.


FantasmicGalaxy:
Before: 
image

After: 
image

Without scaling:
image

dukio:
Already fixed in #781 btw

FantasmicGalaxy:
K, I'll just close then

PR #802, opened 2021-09-09
Fixed API caching in /apiv2

MartinNemi03:
Our API should've been cache-only all along, so this fixes the issue of API requests using external resources.

This bug was created by a commit from me from April 19th, _sorry for that one_.


nstringham (approved):
Can test right now but this looks good

PR #803, opened 2021-09-09
Fixed more things around API caching

MartinNemi03:
There were still some unintended external requests after previous PR, so here we have another PR that attemps to fix it. 

Right as I were adding cacheOnly into other methods used in the API, I went ahead and simplied it using an options object, that contains some things the methods use for caching and debug stuff.

_Let's hope it works this time.._


dukio (approved):
I can't test right now but looked through it on github and it looks good

PR #808, opened 2021-09-10
add divine between mythic and supreme with color §b

### LINKED ISSUE #799: Upcoming new rarity: Divine
Ref: https://hypixel.net/threads/psa-upcoming-gemstone-changes.4523454/page-16#post-32716374

> 1/ The rarity change on the Divan's Drill was uncalled for, so the default rarity will remain Mythic.
> However, we will officialize a rarity above Mythic, called Divine, with its own reforge stats. This rarity will be kept exclusive to items of the caliber of the Divan’s Drill.
> Divine will be added as an AH filter.
> Enchanted Books cannot be Divine.

nstringham:
Don't merge this PR until we can confirm this is how divine works in-game

fixes #799


dukio:
https://github.com/SkyCryptWebsite/SkyCrypt/blob/master/src/lib.js#L641 (631-650) should also be updated to do something similar to SPECIAL and VERY SPECIAl for DIVIDE and SUPREME, as explained in #799

nstringham:
> https://github.com/SkyCryptWebsite/SkyCrypt/blob/master/src/lib.js#L641 (631-650) should also be updated to do something similar to SPECIAL and VERY SPECIAl for DIVIDE and SUPREME, as explained in #799
> 
> (didn't test, can't test, but I image it will have the same issue SPECIAL and VERY SPECIAL have)

I think line 641 might only be required because very special has a space in it but I'm not sure I'll take a closer look

nstringham:
Do we have a profile for testing yet?

dukio:
> Do we have a profile for testing yet?

Can't find any, if I do i'll let u know

nstringham:
Marking this as non-draft but don't merge until someone has tested it

dukio:
http://localhost:32464/stats/DeathStreeks has a DIVINE Divan's Drill but can't be tested cause of the "color" error (fixed in another merged PR), @nstringham can you upstream this?

nstringham:
looks good 
image

dukio:
(about the MISSING VALUES, fixing in another PR right now, it's not related to this PR)

👍

Issue #730, opened 2021-08-11
Unmaintainable joke code


nstringham:
The joke added in https://github.com/SkyCryptWebsite/SkyCrypt/commit/8f588e1c3c9ec9dc7c64d52c5abd8cbcda1fa099 is hard-coded markup Wich means duplicate code Wich means more difficult to maintain code.

metalcupcake5:
removed in #809

PR #809, opened 2021-09-12
Remove hardcoded minion count

metalcupcake5:
wowo metal actually contributing

this pr edits the `minions_max_uniques` constant to automatically calculate how many unique minions a user can craft
it also has a fix for #730

### DIFF
--- a/src/constants/minions.js
+++ b/src/constants/minions.js
@@ -274,7 +274,6 @@ export const minions = {
     name: "Voidling",
     type: "combat",
     head: "/head/3a851ed2ce5c2c0523af772d206d9555e2e1383ec87946e6ff4c51186e29ef7f",
-    tiers: 12,
   },
   HARD_STONE: {
     name: "Hard Stone",
--- a/src/constants/minions.js
+++ b/src/constants/minions.js
@@ -25,7 +25,7 @@ export const minion_slots = {
   650: 26,
 };
 
-export const minions_max_uniques = 631;
+export let minions_max_uniques = 0;
 export const minions_max_slots = 25; // From unique tiers (excludes community shop upgrades)
 
 export const minions = {
@@ -274,6 +274,7 @@ export const minions = {
     name: "Voidling",
     type: "combat",
     head: "/head/3a851ed2ce5c2c0523af772d206d9555e2e1383ec87946e6ff4c51186e29ef7f",
+    tiers: 12,
   },
   HARD_STONE: {
     name: "Hard Stone",
@@ -282,3 +283,7 @@ export const minions = {
     tiers: 12,
   },
 };
+
+for(const minion in minions){
+  minions_max_uniques += minions[minion].tiers || 11;
+}
--- a/src/constants/minions.js
+++ b/src/constants/minions.js
@@ -284,6 +284,6 @@ export const minions = {
   },
 };
 
-for(const minion in minions){
+for (const minion in minions) {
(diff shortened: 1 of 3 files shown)


src/constants/minions.js
```diff
@@ -274,6 +274,7 @@ export const minions = {
     name: "Voidling",
     type: "combat",
     head: "/head/3a851ed2ce5c2c0523af772d206d9555e2e1383ec87946e6ff4c51186e29ef7f",
+    tiers: 12,
```
dukio, line 277: Voidling has 11 tiers, in game crafted generators is bugged

views/stats.ejs
```diff
@@ -110,6 +110,10 @@ const slayerInfo = {
   enderman: {
     boss: 'Voidgloom Seraph',
     head: '/head/1b09a3752510e914b0bdc9096b392bb359f7a8e8a9566a02e7f66faff8d6f89e',
+  },
+  slime: {
```
dukio, line 114: I would just remove slime tbh, only confusing for players

dukio (changes requested):
And the code doesn't seem formatted correctly, all ifs and for are missing a space between if and (

nstringham:
I don't understand why the minion count and the slime slayer are in the same PR but I guess I've done worse

src/lib.js
```diff
@@ -1920,10 +1920,28 @@ export const getStats = async (
       output.slayer_coins_spent.total = output.slayer_coins_spent.total || 0;
     }
 
+    if (profile.uuid == "b44d2d5272dc49c28185b2d6a158d80a") {
```
nstringham, line 1923: I don't like that the slim easter egg is in lib.js because it could potentially affect other things. it would be better to keep it in stats.ejs

src/constants/minions.js
```diff
@@ -25,7 +25,7 @@ export const minion_slots = {
   650: 26,
 };
 
-export const minions_max_uniques = 631;
+export let minions_max_uniques = 0;
```
nstringham, line 28: try to avoid `export let` because it looks like it should be assignable but actually isn't

nstringham (changes requested):
#810

dukio:
Can we just remove the slime slayer? 🙏

metalcupcake5:
wait it didn't let me add to `export const`

metalcupcake5:
nvm i see how u did it now

dukio (approved):
ltgm, didn't test

src/lib.js
```diff
@@ -1923,7 +1923,7 @@ export const getStats = async (
     output.slayer_xp = 0;
 
     for (const slayer in slayers) {
-      if (slayers[slayer]?.level?.currentLevel == undefined) {
+      if (!slayers[slayer]?.level?.currentLevel) {
```
nstringham, line 1926: don't do this. the reason it is `== undefined` is because in this case you are really checking for nullishness not truthiness.

although it might no matter in this case you should avoid checking the truthiness of numbers because `0` is falsie

other things that sometimes work are `x > 0`, `typeof x == "number"`, and `x in obj`

nstringham (approved):
In the future making max minions dynamic and removing slime slayer should have been 2 separate PRs but other than that everything looks good now 👍

src/constants/minions.js
nstringham: @dukio is this resolved?

dukio: @nstringham yes

PR #814, opened 2021-09-13
Gemstone stats updated

### LINKED ISSUE #801: Upcoming gemstone stats changes
Ref: https://hypixel.net/threads/psa-upcoming-gemstone-changes.4523454/page-16#post-32716374

> 7/ Slight buffs to how many stats the Sapphire, Ruby and Amethyst gemstone slots provide.

If the stats will be in the lore, the whole gemstone stats parser thing might be useless and might be removed, to evaluate once the update drops

Also see #790

dukio:
Closes #801

- Updated gemstone stats using the ones posted by an admin [here](https://hypixel.net/threads/september-13th-crystal-hollows-patch.4544534/#post-32805668)
- the value "-1" for missing value is no longer useful so I removed it
- updated main comment

### DIFF
--- a/src/constants/gemstones.js
+++ b/src/constants/gemstones.js
@@ -1,8 +1,7 @@
 /*
  * gemstone stats values:
- * - int: exact value tested by dukioooo on 9 august 2021
+ * - int: exact value
  * - null: value doesn't exist, gem can't be applied to any item of that rarity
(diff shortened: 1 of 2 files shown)


nstringham (approved):
Looks good didn't test

PR #815, opened 2021-09-13
New gemstone slots support (defensive, combat, mining)

### LINKED ISSUE #790: Upcoming gemstone slots parser will require an update
REF1: https://hypixel.net/threads/psa-upcoming-gemstone-changes.4523454/
REF2: https://docs.google.com/spreadsheets/d/1GAVyod8Xfu0rNh69DPg-48IF98VanhDlt8uxlZJRarc/edit#gid=0

Now gemstone slots can be either UNIVERSAL or GEM specific (ex: RUBY), with the upcoming changes there will be new types that we need to support:

- **Defensive** slots are only Ruby and Amethyst (HP and Def)
- **Combat** slots are only Ruby, Amethyst, Sapphire and Jasper (HP, Def, Intel and Strength)
- **Offensive** slots are only Sapphire and Jasper (Intel and Strength)
- **Mining** slots are only Jade, Amber and Topaz (MF, MS and Pristine)

Be aware of this:
> 10/ Add a way to see the gemstones on an item without using addons. (Visible in the stats)

dukio:
Closes #790 

Adds support for Mining, Combat and Defensive gemstone slots types on top of Universal and gem specific ones

Fixes https://sky.shiiyu.moe/stats/snapdoomy `TypeError: Cannot read property 'color' of undefined`

⚠️ I could not find any profile with a MINING slot being used, so that's not tested (and I do not own a X655 or Divan Drill)

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -823,6 +823,36 @@ export function parseItemGems(gems, rarity) {
         gem_type: gems[`${key}_gem`],
         gem_tier: value,
       });
+    } else if (key.startsWith("DEFENSIVE_")) {
+      if (key.endsWith("_gem")) {
+        continue;
+      }
+      parsed.push({
+        slot_type: "DEFENSIVE",
+        slot_number: +key.split("_")[1],
+        gem_type: gems[`${key}_gem`],
+        gem_tier: value,
+      });
+    } else if (key.startsWith("COMBAT_")) {
+      if (key.endsWith("_gem")) {
+        continue;
+      }
+      parsed.push({
+        slot_type: "COMBAT",
+        slot_number: +key.split("_")[1],
+        gem_type: gems[`${key}_gem`],
+        gem_tier: value,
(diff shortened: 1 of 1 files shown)


nstringham:
what does this do?
```js
if (key.endsWith("_gem")) {
  continue;
}
```

dukio:
> what does this do?
> 
> ```js
> if (key.endsWith("_gem")) {
>   continue;
> }
> ```

Basically if an item has a normal gem, it will have only have 
```
ExtraAttributes.gemstones.RUBY_0: 'FINE'
``` 

but if the item has a special slot (universal, mining, defensive or combat) it will have:
```
ExtraAttributes.gemstones.COMBAT_0: 'FINE',
ExtraAttributes.gemstones.COMBAT_0_gem: 'RUBY'
```

so while looping through it we can just ignore the `_gem` key

PR #821, opened 2021-09-14
Quick claw

### LINKED ISSUE #792: Upcoming new pet item (mining speed)
REF: https://hypixel.net/threads/psa-upcoming-gemstone-changes.4523454/

image

dukio:
Support for new Quick Claw pet item:
image

pet items now can have a new property `statsPerLevel` to make this new pet item work

image

Closes #792

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -2897,22 +2897,16 @@ export async function getPets(profile) {
       let heldItemObj = await db.collection("items").findOne({ id: heldItem });
 
       if (heldItem in constants.pet_items) {
-        if ("stats" in constants.pet_items[heldItem]) {
-          for (const stat in constants.pet_items[heldItem].stats) {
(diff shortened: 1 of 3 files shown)


src/lib.js
```diff
@@ -2902,6 +2902,12 @@ export async function getPets(profile) {
             pet.stats[stat] = (pet.stats[stat] || 0) + constants.pet_items[heldItem].stats[stat];
           }
         }
+        if ("statsPerLevel" in constants.pet_items[heldItem]) {
+          for (const stat in constants.pet_items[heldItem].statsPerLevel) {
```
nstringham, line 2906: Can't you just do `for (const stat in constants.pet_items[heldItem]?.statsPerLevel) {` instead of the if statement?

dukio: Didn't think of that, updated that and the other 2 similar IFs

PR #823, opened 2021-09-14
fix: parseItemGems refactor

dukio:
Fixes `TypeError: Cannot read property 'color' of undefined` for:
- https://sky.shiiyu.moe/stats/TheDarkKillerGR
- https://sky.shiiyu.moe/stats/15h

Basically yesterday I didn't knew `OFFENSIVE` was another type of special slot, but I also noticed there's a new key for only the new items in `ExtraAttributes.gemstones.slots_unlocked` that needs to be ignored while parsing the gems.

So I just went ahead and refactored the function a bit, now it's much more cleaner and should also give a straightforward error message in case new unknown keys appear.

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -811,55 +811,37 @@ export function generateUUID() {
  */
 export function parseItemGems(gems, rarity) {
   /** @type {Gem[]} */
+
+  const slots = {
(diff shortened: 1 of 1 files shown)


nstringham (approved):
initial glance over looks good. I like the reduction in redundant code.

PR #826, opened 2021-09-15
Add a link at the top to Hellcastle's gofundme

FantasmicGalaxy:
Could not figure out what to say in the description, this is the best I could think of


nstringham:
screenshot? this is a UI change

PR #827, opened 2021-09-15
Fallback divine gemstone stats (feedback?)

dukio:
This temporarily solves the issue, but I would like some feedback.

Basically in game there's some DIVINE rarity items that aren't supposed to exist, mostly dungeon armor like Necron or Shadow Assassin, example:

https://sky.shiiyu.moe/stats/Bucklez/Coconut#Armor
image

What I think happened is that admins didn't expect/know this items existed so they didn't code in the values for DIVINE stats in gemstone that can't be applied to Divan's Drill (the only item that should have DIVINE rarity as of right now).
In game that item gems use the value as if the item was a a mythic rarity, so as a temporarily solution I coded that if the item is DIVINE and we are missing the gemstone stat value it tries to fallback to MYTHIC.

What do you think?

Instead of doing this we could just hardcode the "fake/temp" values where they are missing, so like Flawless Sapphire will give the same stats in both mythic and divine. But doing this will probably mean that in the future we might not be sure which values were added temporarily and which are the definitive ones... with the risk of forgetting to update some.

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -875,7 +875,13 @@ export function generateGemLore(type, tier, rarity) {
     const gemstone_stats = constants.gemstones[type.toUpperCase()]?.stats?.[tier.toUpperCase()];
     if (gemstone_stats) {
       Object.keys(gemstone_stats).forEach((stat) => {
-        const stat_value = gemstone_stats[stat][rarityNameToInt(rarity)];
+        let stat_value = gemstone_stats[stat][rarityNameToInt(rarity)];
+
+        // Fallback since skyblock devs didn't code all gemstone stats for divine rarity yet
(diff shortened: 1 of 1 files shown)


dukio:
This https://discord.com/channels/738971489411399761/739600697057607783/888100138822799460 confirms my theory

metalcupcake5:
i think the fallback would work fine. do you think admins would revert these items?

metalcupcake5 (approved):

dukio:
> i think the fallback would work fine. do you think admins would revert these items?

Not sure honestly, like they didn't revert exotics yet... but they reverted started warden helmets recently so who knows.

MartinNemi03 (approved):

PR #843, opened 2021-09-24
Modified Connections from 10 to 6

WarpWing:


WarpWing:
> what does this do?

I was reviewing the nginx configurations with Martin in DMs and this basically limits the page connections(not API) from 10 concurrent to 6 concurrent at the max per host.

nstringham:
What does per host mean?

WarpWing:
> What does per host mean?

Per user basically.

nstringham:
> > What does per host mean?
> 
> Per user basically.

Per IP?

WarpWing:
> > > What does per host mean?
> > 
> > 
> > Per user basically.
> 
> Per IP?

Yes

nstringham:
What is the reasoning for restricting it to 6?
Are we just thinking the there will never be more than 6 users on the same WiFi at the same time?

WarpWing:
> What is the reasoning for restricting it to 6?
> Are we just thinking the there will never be more than 6 users on the same WiFi at the same time?

To prevent any sort of spam attack or over connection to SkyCrypt

nstringham:
> > What is the reasoning for restricting it to 6?
> > Are we just thinking the there will never be more than 6 users on the same WiFi at the same time?
> 
> To prevent any sort of spam attack or over connection to SkyCrypt

Yes I get that but like why isn't 10 sufficient why 6?

WarpWing:
> > > What is the reasoning for restricting it to 6?
> > > Are we just thinking the there will never be more than 6 users on the same WiFi at the same time?
> > 
> > 
> > To prevent any sort of spam attack or over connection to SkyCrypt
> 
> Yes I get that but like why isn't 10 sufficient why 6?

I just ballparked a number but we certainly don't need 10. I think 5-6 is fine. I talked to Martin about this in DMs, he seemed to agree on the idea and I think 6 is a reasonable number. 
image

nstringham:
I mean a lan part of 7 people who all have SkyCrypt open in the background would be enough to break it

Do we at least have a useful error message when the limit is exceeded?

dukio:
> I mean a lan party of 7 people who all have SkyCrypt open in the background would be enough to break it
> 
> Do we at least have a useful error message when the limit is exceeded?

Is this actually the case? once a page is loaded the connection is closed...

From my understanding this just limits the number of pages you can load at the same time... and honestly even 6 is very generous

WarpWing:
> > I mean a lan party of 7 people who all have SkyCrypt open in the background would be enough to break it
> > Do we at least have a useful error message when the limit is exceeded?
> 
> Is this actually the case? once a page is loaded the connection is closed...
> 
> From my understanding this just limits the number of pages you can load at the same time... and honestly even 6 is very generous

Yeah 6 is pretty generous, but I was thinking it's somewhat of a starting point. I was considering 4-5 but I'm not too sure about the scale of our users in order to figure that limit out. Maybe change it to 4 or 5?

WarpWing:
> I mean a lan party of 7 people who all have SkyCrypt open in the background would be enough to break it
> 
> Do we at least have a useful error message when the limit is exceeded?

I mean that's a good point. We should tell the users on some level that their connection is exceeded.

nstringham:
> Is this actually the case? once a page is loaded the connection is closed...
> 
> From my understanding this just limits the number of pages you can load at the same time... and honestly even 6 is very generous

I am pretty sure the browser keeps the connection open for several minutes so that it doesn't need to reopen the connection when the user navigates

dukio:
Was this already "merged" (as in pushed to the server)?

Right now if I try to open 12 profiles all at once, 6 will load (or be loading) and the rest will return 404 instantly... as soon as a profile load I can load one more profile...

If this is the case, yeah I think we need a better error/page than default nginx 404... but also what @nstringham said about browsers keeping connection open for several minutes doesn't apply

To reproduce simply add a 4+ favourites and open in new tab all the profiles in the home page

WarpWing:
> Was this already "merged" (as in pushed to the server)?
> 
> Right now if I try to open 12 profiles all at once, 6 will load (or be loading) and the rest will return 404 instantly... as soon as a profile load I can load one more profile...
> 
> If this is the case, yeah I think we need a better error/page than default nginx 404... but also what @nstringham said about browsers keeping connection open for several minutes doesn't apply
> 
> To reproduce simply add a 4+ favourites and open in new tab all the profiles in the home page

This is weird. It should allow at least 10 pages to load before a 404 occurs. Is there something on another application level that's blocking more than 10 simultaneous connections?

nstringham:
> This is weird. It should allow at least 10 pages to load before a 404 occurs. Is there something on another application level that's blocking more than 10 simultaneous connections? Is it possibly Cloudflare?

I don't think it's Cloudflare because it is an nginx error

nstringham:
I did a test where I loaded SkyCrypt in 12 browsers on 5 devices and it seemed to work except for one time when there was a generic network error on one browser and one time where windows crashed

WarpWing:
> I did a test where I loaded SkyCrypt in 12 browsers on 5 devices and it seemed to work except for one time when there was a generic network error on one browser and one time where windows crashed

Did you use the SkyCrypt Nginx Configuration? This connection limit is on the Nginx level. Just want to make sure because it should be closing.

nstringham:
> > I did a test where I loaded SkyCrypt in 12 browsers on 5 devices and it seemed to work except for one time when there was a generic network error on one browser and one time where windows crashed
> 
> Did you use the SkyCrypt Nginx Configuration? This connection limit is on the Nginx level. Just want to make sure because it should be closing.

I used sky.shiiyu.moe because I don't have nginx set up locally.

dukio (approved):

PR #850, opened 2021-09-27
current forge processes

metalcupcake5:
this pr adds a display for the current items in the forge
currently it shows them in a text form, such as the examples below
image
image
kinda working on a version that shows items, has kinda messy code so idk if i'll follow through with it


metalcupcake5:
oops i forgot to lint the files 😹

nstringham:
"in in"

dukio:
I would make it more compact, like:
image


```html
          mining.forge.processes.forEach(process => { %>
            <div style="display: flex; gap: 8px; align-items: center;">
              <p class="stat-name" style="margin: 0">Slot <%= process.slot %>:</p>
              <span data-tippy-content='Ending at <local-time timestamp="<%= process.timeFinished %>"></local-time>'><span class="stat-raw-values"><%= process.name %> - ending <%= process.timeFinishedText %></span></span>
            </div>
          <% });
```
(ofc with proper classes instead of inline styles)

dukio:
image

Also it should say `ended` instead of `ending in xx time` for completed items

nstringham:
I would put the tippy on the moment result only
image
```html
<span class="stat-raw-values"><%= process.name %> - ending <span data-tippy-content='<local-time timestamp="<%= process.timeFinished %>"></local-time>'><%= process.timeFinishedText %></span></span>
```

src/lib.js
```diff
@@ -3491,6 +3493,44 @@ export async function getDungeons(userProfile, hypixelProfile) {
...
+        forgeItem.name = item.id == "PET" ? "[Lvl 1] Ammonite" : dbObject ? dbObject.name : item.id;
+        const timeFinished = item.startTime + forgeTime;
+        forgeItem.timeFinished = timeFinished;
+        forgeItem.timeFinishedText = moment(timeFinished).fromNow();
+      }
+      processes.push(forgeItem);
+    }
+    output.processes = processes;
+
+    return output;
```
nstringham, line 3528: why is the output returned both inside and outside the if?

PR #905, opened 2021-10-24
Remove unused npm dependencies

nstringham:
we where requiring some things that we never use so I removed them

I also removed `lodash.escaperegexp` because we already use `lodash`


nstringham:
Waiting on #909

metalcupcake5 (approved):

dukio (approved):
lgtm didn't test

PR #907, opened 2021-10-24
Fix eslint issues

dukio:
This PR just fixes all eslint issues by:
- Un-deprecating `hasPath()` and `getPath()` (they have been converted to optional chaining everywhere possible)
- Converting typescript type `any` to the correct types (I don't have much experience with this so please check if everything I did is correct)
- Converting `navigator.platform` (deprecated) to `navigator.userAgentData.platform` and updating the `Navigator` interface in `global.d.ts`
- Removing unused vars
- Changing the type of an unused var from `any` to `unknown`, so we don't get useless error (it's the first argument of a function where we need the second, so we can't just remove it)
- In stats-defer making sure `item.tag.display` exists before checking and getting the Lore (eslint didn't gave error but vscode did)

### DIFF
--- a/public/resources/ts/stats-defer.ts
+++ b/public/resources/ts/stats-defer.ts
@@ -7,7 +7,8 @@ import("./elements/inventory-view");
 const favoriteElement = document.querySelector(".favorite") as HTMLButtonElement;
 
 if ("share" in navigator) {
-  const platform = navigator.userAgentData?.platform || "unknown";
+  // eslint-disable-next-line deprecation/deprecation
+  const platform = navigator.platform ?? navigator.userAgentData?.platform ?? "unknown";
   const shareIcon = platform.match(/(Mac|iPhone|iPod|iPad)/i)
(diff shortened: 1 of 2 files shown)


public/resources/ts/stats-defer.ts
```diff
@@ -7,7 +7,9 @@ import("./elements/inventory-view");
 const favoriteElement = document.querySelector(".favorite") as HTMLButtonElement;
 
 if ("share" in navigator) {
-  const shareIcon = navigator.platform.match(/(Mac|iPhone|iPod|iPad)/i)
+  // eslint-disable-next-line deprecation/deprecation
+  const platform = navigator.platform ?? navigator.userAgentData?.platform ?? "unknown";
```
nstringham, line 11: we could import this from `common.ts` but this is fine

nstringham (approved):
this all looks good

tested on iOS 15

metalcupcake5:
could `hasPath` just be removed?

PR #909, opened 2021-10-25
replace @tillhub/tableify with custom function

nstringham:
the [@tillhub/tableify](https://github.com/tillhub/tableify) npm package is causing some problems and it doesn't appear to be maintained so this PR replaces it with a custom function that does the same thing

I created this function by starting with the score code from @tillhub/tableify and removing the features we didn't need. then I optimized the code and made it more readable.

I tested this PR with Cookie_Wookie_7/Papaya and everything should be identical to the way it worked before


MartinNemi03 (approved):
LGTM

Issue #597, opened 2021-07-11
External Theme Providers


nstringham:
**Is your suggestion related to a problem? Please describe in detail.**
This suggestion would allow users to have many more themes available to them without the need for SkyCrypt to maintain a overwhelmingly large list of themes

**Describe your suggestion**
Theme creators could host a JSON file containing a SkyCrypt theme at a spastic url. Users would simply paste the url into SkyCrypt and the client would handle fetching validating and applying the theme.

SkyCrypt would store two values in local storage: the JSON of the current theme so that it can be quickly applied on page load, and the url of the current theme to update the JSON with. The client could also store a list of previously used theme URLs to allow you to switch between multiple themes.

**Describe alternatives you've considered**
An alternative would be to allow the user to paste the JSON of the desired theme in directly however the disadvantage of this is that it would not allow themr authors to remotely update the theme as changes happen to the site

**Additional context**
If this feature was as I would likely add an option for SkyCrypt themer users to publish their themes to a URL This would allow SkyCrypt thememer users to use custom themes on devices that do not run desktop Chrome such as iPhone and Android

FantasmicGalaxy:
Sounds like a good idea, also gives us a good reason to expand the theme options, I'll just have to look into if there are any problems with XSS

nstringham:
> I'll just have to look into if there are any problems with XSS

I think [Response.json()](https://developer.mozilla.org/en-US/docs/Web/API/Response/json) is good enough to sanitize for us as long as we never use [eval()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval)

PR #917, opened 2021-11-01
Theme JSON from URL

nstringham:
this PR changes the client so that it gets themes from URLs as JSON.

it also renders `#theme-box` on the client now

this PR also adds the ability for other websites to temporarily change the theme. for an example of this behavior check out the future [skycrypt themer website](https://skycrypt-themer.web.app/)

this is the first step to implementing #597

this PR adds the possibility for 3rd party themes provided but does not add a GUI for adding them.

external themes are loaded from `customThemeUrls` in localStorage as a JSON array.
This functionality can be tested by running this command in the console and then reloading the page
```js
localStorage.setItem("customThemeUrls",'["/resources/themes/warpwing.json"]')
```


WarpWing:
@nstringham In regards to the PR in a security lens in orders, the reason why CodeQL is complaining is due to the plain text being written in JSON unverified. To mitigate this, I recommend validating the JSON or finding some sort of sanitization to prevent the JSON from possibly being exploited(check JSON against common link patterns or writing a CSP(Content Security Policy) to restrict what can be loaded. However, I would recommend not even allowing possible unsanitized JSON even be loaded in the first place. Just sanitize it and escape it to prevent any malicious data from even being allowed to be check against in the first place.

WarpWing (approved):

metalcupcake5 (approved):
did not test

MartinNemi03 (approved):

PR #940, opened 2021-11-06
remove stuff from global

nstringham:
currently SkyCrypt gives this error
image
this PR fixes that by removing stuff that doesn't need to be on the global from the global


dukio (approved):
lgtm

metalcupcake5 (approved):

PR #943, opened 2021-11-08
forge processes unknown and scrolls

### LINKED ISSUE #942: New forge items/recipes to add
> Fixes to warps:
> "/warp forge" now exists along side "/warpforge" and now actually requires a Travel Scroll.
> Crystal Hollows now have their own Travel Scroll.
> Both the Dwarven Forge and Crystal Hollows Travel Scrolls are obtained from the Forge as an item cast.

dukio:
This PR does 2 things:
1. Closes #942 by adding the times for the new recipes
2. Changes the HTML so that it only uses 1 `<span>` instead of 2 and also uses the correct `stat-value` class instead of `stat-raw-values` which is correctly vertically aligned (and it's what we usually use everywhere else)
3. Adds a fallback in case of future new items, so the forge processes won't look like this:

old:
image

new:
image

---
image

PS: The name is parsed by items collection so it will get fixed soon, whenever builder's API updates
```js
const dbObject = await db.collection("items").findOne({ id: item.id });
forgeItem.name = item.id == "PET" ? "[Lvl 1] Ammonite" : dbObject ? dbObject.name : item.id;
```

next day edit:
image

there they are

### DIFF
--- a/src/constants/forge.js
+++ b/src/constants/forge.js
@@ -63,6 +63,8 @@ export const forge_times = {
   BOOTS_OF_DIVAN: 23 * 60,
   DIVAN_DRILL: 2 * 24 * 60 + 12 * 60,
   PET: 12 * 24 * 60,
+  FORGE_TRAVEL_SCROLL: 5 * 60,
+  CRYSTAL_HOLLOWS_TRAVEL_SCROLL: 10 * 60,
 };
 
 export const quick_forge_multiplier = {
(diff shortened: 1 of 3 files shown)


nstringham:
Thanks

nstringham (approved):
Looks good didn't test

metalcupcake5 (approved):

PR #998, opened 2021-11-25
Refactor weight systems

nstringham:
this refactor makes it so that code is not duplicated between lily and senither weight

this PR also fixes a problem where an extra space is added between the weight number and the tippy `*`

this PR removes the word "provided" from weight attribution so that it will all fit on one line

this PR also makes the weight round instead of floor because everyone likes bigger numbers :)

image


metalcupcake5 (approved):

dukio (approved):
lgtm didnt test

MartinNemi03 (approved):

PR #1009, opened 2021-12-02
fix: show gauntlet in both weapons and pickaxes

### LINKED ISSUE #858: Gauntlet with combat enchants shows in weapons but not in mining tools
https://sky.shiiyu.moe/stats/AetherSword/Cucumber

The gauntlet should display both in weapons and mining tools, but in this case it only shows in the weapons section.
Maybe it's because it has weapon enchants?

Also
image

metalcupcake5:
fixes #858 
tested on AetherSword/Cucumber
image
image

### DIFF
--- a/src/constants/misc.js
+++ b/src/constants/misc.js
@@ -18,6 +18,7 @@ export const item_types = [
   "chestplate",
   "leggings",
   "boots",
+  "gauntlet",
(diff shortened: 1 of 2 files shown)


dukio (approved):
lgtm didnt test

PR #1029, opened 2021-12-15
fix fix lilyweight

dukio:
basically lilyweights considers all skills already at level 60 so we need to calculate the level faking a maxLevel of 60 for everything... this PR does that

what Nate did in his first fix PR did pretty much the same, but for the entire site... so players were missing a lot of overflow xp on max skills

honestly idk if this is the best way to fix the issue... I never really liked how we handle the parsing the xp to level, overflow xp etc...

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -137,7 +137,7 @@ function getXpByLevel(level, extra = {}) {
   return output;
 }
 
-export function getLevelByXp(xp, extra = {}) {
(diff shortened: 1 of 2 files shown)


metalcupcake5 (approved):

nstringham:
I kinda don't like the way that `getLevelByXp` is now called twice. I'm going to see if I can find a better solution

nstringham:
I think #1031 is a better solution to this problem but if we don't do that the I'll take a closer look at this

PR #1038, opened 2021-12-21
added the speed tiers reached in the races

doej1367:
Screenshot 2021-12-21 at 12-18-31 john_linux (Cucumber)

Description:
Pretty much does exactly what the screenshot shows. 
It adds the completed race speed tiers from the objectives api.

~~Sidenote: 
Converting the objective names into the race names, I didn't know what would be uglier. Coding in the few edge cases that tere are or making a lot of dictionary entries... for now I chose the first option - in hindsight the second option would have been better, but I want to hear your opinion on the implementation first.~~
EDIT: changed it to the second option

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -454,8 +454,14 @@ export function titleCase(string) {
   return split.join(" ");
 }
 
-export function renderRaceTier(tierNumber) {
-  return "●".repeat(tierNumber) + "○".repeat(4 - tierNumber);
+/**
+ * returns a string with 4 dots "●" for completed tiers and "○" for incomplete tiers
+ * @param {number} completeTiers
+ * @returns {string} 4 dots
+ */
+export function renderRaceTier(completeTiers) {
+  const incompleteTiers = Math.max(0, 4 - completeTiers);
+  return "●".repeat(completeTiers) + "○".repeat(incompleteTiers);
 }
 
 /**
--- a/src/helper.js
+++ b/src/helper.js
@@ -455,14 +455,7 @@ export function titleCase(string) {
 }
 
 export function renderRaceTier(tierNumber) {
-  let res = "";
-  for (let i = 0; i < tierNumber; i++) {
-    res += "●";
-  }
-  for (let i = 0; i < 4 - tierNumber; i++) {
-    res += "○";
-  }
-  return res;
+  return "●".repeat(tierNumber) + "○".repeat(4 - tierNumber);
 }
 
 /**
--- a/src/helper.js
+++ b/src/helper.js
@@ -454,6 +454,17 @@ export function titleCase(string) {
   return split.join(" ");
 }
 
+export function renderRaceTier(tierNumber) {
+  let res = "";
+  for (let i = 0; i < tierNumber; i++) {
+    res += "●";
+  }
+  for (let i = 0; i < 4 - tierNumber; i++) {
+    res += "○";
+  }
+  return res;
+}
+
 /**
  * checks whether a string should be proceeded by a or by an
  * @param {string} string
--- a/public/resources/scss/shared.scss
+++ b/public/resources/scss/shared.scss
@@ -2861,8 +2861,11 @@ inventory-view {
   padding-bottom: 30px;
 }
 
-.tier-completion {
-  float: right;
+.narrow-info-container .narrow-info-flexsb {
+  display: flex;
+  flex-flow: row nowrap;
+  align-items: flex-start;
+  justify-content: space-between;
 }
 
 .slayer {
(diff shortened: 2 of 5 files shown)


dukio:
Gave a quick look and seems good to me, only thing I noticed that I would change is the float to render the raceTier...

Avoid floats at all cost...

Will give it a better look after work and also test it

doej1367:
> only thing I noticed that I would change is the float to render the raceTier

The reason I did that is that I don't know a different way to align it to the right. Is there a better way to do that?

> Will give it a better look after work and also test it

Thx :)

nstringham:
> The reason I did that is that I don't know a different way to align it to the right. Is there a better way to do that?

You should probably use grid or maybe flex

nstringham:
It seems kind of weird for the dots to be on the right side of the dungeon races but the left side of the other races. What does it look like if the dots are on the right side of all the races?

doej1367:
> What does it look like if the dots are on the right side of all the races?

Then the dots for the first three get to the very right and they look really out of place. Putting them all on the left side however would look like this:
Screenshot 2021-12-21 at 19-55-25 john_linux (Cucumber)
Other brackets or indicator symbols would also be an option...

nstringham:
what do you all think about this?
image

doej1367:
> what do you all think about this? image

Looks great. That's exactly what I wanted to do at first, but I'm a newb in css so I couldn't realize it. The second thing I wanted to do was some kind of border around the first few so that the dots aren't in mid air like this (again, I don't have the css skills to do that, yet).

dukio:
> what do you all think about this? image

Looks good but I agree the first 2 look kind of "in the air"

Maybe put them inside another box called "other races"?

src/helper.js
```diff
@@ -454,6 +454,17 @@ export function titleCase(string) {
...
 
+export function renderRaceTier(tierNumber) {
+  let res = "";
+  for (let i = 0; i < tierNumber; i++) {
+    res += "●";
+  }
+  for (let i = 0; i < 4 - tierNumber; i++) {
+    res += "○";
+  }
+  return res;
```
dukio, line 465: This could be simplified with:

```js
return "●".repeat(tierNumber) + "○".repeat(Math.max(0, 4-tierNumber));
```

dukio (commented):
Little change I would do to simplify the code a bit

src/helper.js
nstringham: lol I already did this

nstringham: wait what is the `Math.max(0` for?

nstringham: isn't is already positive?

dukio: @nstringham in case in the future they will add a tier above 4... so it won't go `.repeat(-1)` and give error

dukio: Ofc we would need to update the "4" of "4-tierNumber" to fix it... but in the meanwhile the profiles won't load

nstringham:
> Maybe put them inside another box called "other races"?

like this?
image

dukio:
> > Maybe put them inside another box called "other races"?
> 
> like this? image

If you put them as last box in the list will the box still be super tall? Right now it's a lot of empty space 😕 

my idea is if it wraps to the next line it would only use the needed vertical space... or at least I hope

nstringham:
I like this better image

dukio:
> I like this better image

lol perfect 👍

nstringham (changes requested):
I have made some changes in https://github.com/doej1367/SkyCrypt/pull/48

src/helper.js
doej1367: `return "●".repeat(tierNumber) + "○".repeat(Math.max(0, 4-tierNumber));`

So I learnt thats a thing... didn't know about the repeat function until today. thx :)

nstringham (changes requested):
I have added Math.max with @dukio's advice and changed `||` to `??` in https://github.com/doej1367/SkyCrypt/pull/49

doej1367:
> changed `||` to `??`

oh yea that was what I ment to do - sorry, my bad

nstringham:
> Might also not be the only one in this project. What about this one for example?

yeah there are a bunch I have bean fixing them a few at a time when they come up. I don't want to just find replace them because there are some cases where it actually is appropriate to use `||`.

doej1367:
> I don't want to just find replace them because there are some cases where it actually is appropriate to use `||`.

yea, probably better taht way. Already saw all those hundreds of correect logical ORs over these files.

MartinNemi03 (approved):
LGTM!

metalcupcake5 (approved):

nstringham (approved):
this looks good from a UI perspective.

I didn't verify that the data is accurate

dukio:
uhm we keeping that float? 👀 😟

dukio:
let's remove that float? https://github.com/doej1367/SkyCrypt/pull/51

PR #1045, opened 2021-12-25
enderman slayer bonus stats

metalcupcake5:
adds the bonus stats that enderman slayer gives
Before(LeaPhant/Grapes, currently eman 4):
image
After:
image


MartinNemi03 (approved):
LGTM, didn't test

PR #1050, opened 2021-12-29
always throw Errors not strings

nstringham:
it is generally good practice to throw instances of `Error` and not strings. This PR makes this change and adds an eslint rule to prevent regression.


MartinNemi03 (approved):

dukio (approved):
lgtm, didn't test

metalcupcake5 (approved):
look good no test

PR #1051, opened 2021-12-29
Fix forge times

metalcupcake5:
fixes titanium artifact and sapphire polished drill engine times, adds diamonite forge time


nstringham (approved):
looks good didn't test

PR #1053, opened 2021-12-30
essence refactor

dukio:
image

Basically now there's no more hardcoded variables/names/stuff in `lib.js` or `stats.ejs`, it's all handled in `constants/dungeons`

Also using chips 🤷


views/stats.ejs
```diff
@@ -2067,19 +2067,24 @@ const metaDescription = getMetaDescription()
           <% } %>
           </div>
         <% } %>
+
+        <% if (calculated.dungeons.essence) { %>
+          <p class="stat-sub-header">Essence</p>
```
metalcupcake5, line 2072: could you re-add the essence icon

nstringham: It might look cleaner without the icon but I would be interested to see what it looks like

nstringham (approved):
This is a good idea from a technical standpoint but I'm not 100% I like the chips better. I would be interested to see what other people think. I could go either way really

views/stats.ejs
dukio: I think it's more in line with the rest of the dungeon section without it... but yeah I can try to re-add it and send u a preview...

But not before a couple days, I'll be away for new year

dukio:
> This is a good idea from a technical standpoint but I'm not 100% I like the chips better. I would be interested to see what other people think. I could go either way really

I think the vertical design used a lot of vertical space while also wasting all the horizontal space...

Also more dungeons will eventually come (and admins already talked about adding essence to other aspects of the game) so we need to think for the future and allow the element to be espandibile... If they add 20 more essence types the vertical design would waste an insane amount of space :/

dukio:
# Without icon:
image

# With icon:
image

---
I can see that with the icon is more colorful... but Boss Collection doesn't have one so I feel like either we give both an icon or none...

nstringham:
> but Boss Collection doesn't have one so I feel like either we give both an icon or none...

Agree

dukio:
closing cause git issues... remaking this

PR #1076, opened 2022-01-09
fix livid dungeon colelction

metalcupcake5:
changes sa chestplate in livid collection to a different name so it shows as claimed
before([VedatsGT](https://sky.shiiyu.moe/stats/VedatsGT/Kiwi#Slayer))
image
after
image


dukio (approved):
lgtm didn't test

PR #1104, opened 2022-01-30
Refork worker processes when they exit

MartinNemi03:
This PR basically does what is said in the title

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -13,11 +13,12 @@ if (cluster.isPrimary) {
   }
 
   cluster.on("exit", (w, c, s) => {
-    let ts = new Date().getTime();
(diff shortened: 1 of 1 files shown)


dukio:
Honestly no idea what these things are so I can't really review that aspect...

But that `new Date().getTime()` can be replaced with `Date.now()` which makes you save 8 whole characters 🤷‍♂️

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now

MartinNemi03:
> Honestly no idea what these things are so I can't really review that aspect...
> 
> But that `new Date().getTime()` can be replaced with `Date.now()` which makes you save 8 whole characters 🤷‍♂️
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now

Oh nice, thanks for recommending that. Guess I can abandon that 'ts' variable right away.

PR #1126, opened 2022-02-03
use SKYCRYPT_PORT as port if defined

nstringham:
if an environment variable called `SKYCRYPT_PORT` is defined it will be used as the port number otherwise `32464` will be used as usual


MartinNemi03 (approved):
lgtm, at least it's here now

PR #1127, opened 2022-02-04
fix maximum fairy souls on stranded and remove unnecessary values for misc

Shiiyu:
Changes maximum fairy souls on stranded to 3.

### DIFF
--- a/src/constants/misc.js
+++ b/src/constants/misc.js
@@ -254,24 +254,9 @@ export const ranks = {
     tag: "BUILD TEAM",
   },
 
-  MODERATOR: {
-    color: "2",
-    tag: "MOD",
-  },
-
   GAME_MASTER: {
     color: "2",
     tag: "GM",
-  },
-
-  HELPER: {
-    color: "9",
-    tag: "HELPER",
-  },
-
-  "JR HELPER": {
-    color: "9",
-    tag: "JR HELPER",
   },
 
   YOUTUBER: {
--- a/src/constants/misc.js
+++ b/src/constants/misc.js
@@ -423,4 +423,7 @@ export const increase_most_stats_exclude = [
   "pristine",
 ];
 
-export const MAX_SOULS = 228;
+export const MAX_SOULS = {
+  normal: 228,
+  stranded: 3
+};
--- a/src/constants/misc.js
+++ b/src/constants/misc.js
@@ -410,5 +410,5 @@ export const increase_most_stats_exclude = [
 
 export const MAX_SOULS = {
   normal: 228,
-  stranded: 3
+  stranded: 3,
 };
--- a/src/constants/misc.js
+++ b/src/constants/misc.js
@@ -422,3 +422,5 @@ export const increase_most_stats_exclude = [
   "foraging_fortune",
   "pristine",
 ];
+
+export const MAX_SOULS = 228;
--- a/src/lib.js
+++ b/src/lib.js
@@ -1710,7 +1710,7 @@ export const getStats = async (
   output.fairy_souls = {
     collected: userProfile.fairy_souls_collected,
     total: totalSouls,
-    progress: (userProfile.fairy_souls_collected / totalSouls),
+    progress: userProfile.fairy_souls_collected / totalSouls,
   };
 
   const levelCaps = {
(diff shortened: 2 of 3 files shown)


metalcupcake5 (approved):
look god no test

dukio:
This shouldn't be in stats.ejs imo, that's just for displaying stats... 

I would implement the same logic in lib.js#L1710

```js
  output.fairy_souls = {
    collected: userProfile.fairy_souls_collected,
    total: MAX_SOULS,
    progress: Math.min(userProfile.fairy_souls_collected / MAX_SOULS, 1),
  };
```

---

Edit... hold on why the frick is this in lib.js
image

this should be in constants... !??!

nstringham (changes requested):
Similar to dukio, I think we should be changing the value of `calculated.fairy_souls.total` instead of putting logic in the template.

nstringham (approved):

src/constants/misc.js
```diff
@@ -422,3 +422,5 @@ export const increase_most_stats_exclude = [
   "foraging_fortune",
   "pristine",
 ];
+
+export const MAX_SOULS = 228;
```
dukio: This could be:

```js
export const fairy_souls: {
  max: {
    normal: 228,
    stranded: 3,
  }
}
```
Because right now the "3" for stranded is still hardcoded in lib.js

So in lib.js you can do:
```js
const totalSouls = profile.game_mode === "island"
  ? constants.fairy_souls.max.stranded
  : constants.fairy_souls.max.normal;
```

src/lib.js
```diff
@@ -1704,11 +1703,12 @@ export const getStats = async (
...
     }
   }
+  const totalSouls = profile.game_mode === "island" ? 3 : constants.MAX_SOULS;
 
   output.fairy_souls = {
     collected: userProfile.fairy_souls_collected,
-    total: MAX_SOULS,
-    progress: Math.min(userProfile.fairy_souls_collected / MAX_SOULS, 1),
+    total: totalSouls,
+    progress: Math.min(userProfile.fairy_souls_collected / totalSouls, 1),
```
dukio: I noticed we are calculating progress but not displaying it 🤔 
Maybe it should be a tooltip on the image thing

Btw idk if the `Math.min()` is still useful cause they fixed having more than max fairy souls by giving each fairy soul an uuid. So if the progress is > 1 it's our fault for having an outdated `constants.fairy_souls` max value

Shiiyu:
If the only request change is the text format on fairy souls tippy I will not request a review again.

src/lib.js
```diff
@@ -1704,11 +1703,14 @@ export const getStats = async (
...
+  const totalSouls = profile.game_mode === "island"
+    ? constants.MAX_SOULS.stranded
+    : constants.MAX_SOULS.normal;
 
   output.fairy_souls = {
     collected: userProfile.fairy_souls_collected,
-    total: MAX_SOULS,
-    progress: Math.min(userProfile.fairy_souls_collected / MAX_SOULS, 1),
+    total: totalSouls,
+    progress: (userProfile.fairy_souls_collected / totalSouls),
```
dukio: Looks good but it's not formatted (prettier)... the parenthesis here are useless

src/constants/misc.js
```diff
@@ -254,26 +254,11 @@ export const ranks = {
     tag: "BUILD TEAM",
   },
 
-  MODERATOR: {
-    color: "2",
-    tag: "MOD",
-  },
-
```
dukio, line 261: This is not related to this PR though...

src/constants/misc.js
```diff
@@ -254,26 +254,11 @@ export const ranks = {
...
-  HELPER: {
-    color: "9",
-    tag: "HELPER",
-  },
-
-  "JR HELPER": {
-    color: "9",
-    tag: "JR HELPER",
-  },
-
```
dukio, line 276: This is not related to this PR though...

dukio:
Apart from the little things I commented on looks good to me... (will approve when checks pass)

src/lib.js
Shiiyu: image
image

src/constants/misc.js
Shiiyu: image
image

nstringham (changes requested):
I don't see why `ranks` is modified in this PR and you need to fix the prettier issue but other than that this looks good

PR #1129, opened 2022-02-06
CD SSH tweeks

nstringham:
this PR makes deploy job fail if deploy script has error


nstringham:
This PR no longer hardcodes the SSH port to 22

PR #1141, opened 2022-02-11
Fixed some HotM tokens

MartinNemi03:
I know this isn't really needed to be fixed, but now there shouldn't be negative HotM tokens in the display.

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -3501,9 +3501,12 @@ export function getMiningCoreData(userProfile) {
 
   const totalTokens = helper.calcHotmTokens(output.tier.level, data.nodes.special_0);
   output.tokens = {
-    total: totalTokens,
+    total: Math.max(totalTokens, (data.tokens_spent || 0) + (data.tokens || 0)),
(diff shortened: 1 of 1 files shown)


dukio:
Personally I would keep the negative number... if people have bugged tokens we should display it.

Also the bugged property if not used should be removed?

dukio:
Also you are using `data.tokens` which is bugged most likely and always returns 0, example for my profile:

image

tokens: 0
tokens_spent: 7

But I have 15 tokens total, so I have 8 available and 7 used:
image
image

MartinNemi03:
> Also you are using `data.tokens` which is bugged most likely and always returns 0, example for my profile:
> 
> image
> 
> tokens: 0 tokens_spent: 7
> 
> But I have 15 tokens total, so I have 8 available and 7 used: image image

That's why I use the higher value of your calculated value and the `data.tokens` thing.
For the bugged value, yeah it's unnecessary, just added because why not.
Idk why I got hyperfocused on this "bug" when it's not even our problem to fix in the first place, so take this PR as you want.

PR #1171, opened 2022-03-02
fix skill icon shine

dukio:
bug:
image

we probably never noticed but the shine on skills was going outside the box... this PR makes the .piece-shine have a border-radius equal to the .skill-icon element, so it doesn't overflow


nstringham:
I think the ideal solution to this problem would be clipping I'll look into it later tonight

PR #1178, opened 2022-03-03
fix lilyweight [B]: remove lilyweight

dukio:
This PR is to fix lilyweight dungeon weight returning NaN, caused by new M7 floor.

This fix simply prevents lily weight from being printed in front end. We will re-enable it once the npm package gets updated with a fix.


nstringham:
It would be nice to display a message explaining what happened instead of just removing it

PR #1193, opened 2022-03-08
Refactor item types and rarity parsing

### LINKED ISSUE #817: constant > misc.js > item_types should include all the new types?
I noticed that https://github.com/SkyCryptWebsite/SkyCrypt/blob/0f97deca5d09257cc5520f6239c5c264c612f634/src/constants/misc.js#L7-L21 is missing a lof of types:
- Gauntlet
- Dungeon ...everything (helmet boots sword bow rod...)
- Fishing Weapon
- probably more I can't think of right now

Should this get updated?

dukio:
The aim of this PR is to refactor the code that handles item types and rarity:
- `item.equipmentType` and `item.type` will be removed and replaced by `item.categories: []` (array of multiple categories, ex: "armor", "helmet")
- while doing this I'll also remove the localized items detection (hypixel doesn't support multilanguage anymore, if it will in the future we can reimplement a better solution to detect multilanguage)
- `item.rarity` now allows "very special" rarity (=> "very_special", underscore replaces spaces to allow it to be used for css classes)
- `output.pickaxe` is not `output.mining_tools`
- `output.hoes` is now `output.farming_tools`
- `output.rods` is now `output.fishing_tools`
- same with `highest_tier_pickaxe|hoe|rod` -> `highest_tier_mining|fishing|farming_tool`

Visually almost nothing has changed:
- `Mining Pickaxes` has become `Mining Tools`, same for Farming
- `Active Pickaxe` is now `Active Tool`, same for Farming
- Since weapons now allow `WAND` type too, now wands will appear in the weapons section

Closes #817


dukio:
I tested this with plenty of profiles, but please do some tests you too!

Example: [MrsDiana (Tomato)](https://sky.shiiyu.moe/stats/MrsDiana/Tomato) was returning an error due the profile not having all storage slots filled with backpacks (now fixed)

dukio:
Closes #817

metalcupcake5:
does including wands in the check catch healing wands from zombie slayer

dukio:
> does including wands in the check catch healing wands from zombie slayer

yep... they are technically wands... hardcoding an exclude for them... idk how I feel about this 😕 
but I have to hardcode (in another future PR) something similar for Pumpkin and Melon Dicer (so they end up in farming tools)... so that PR might include an exclude from weapons for healing wands

metalcupcake5:
got it 👍

metalcupcake5 (approved):

MartinNemi03 (approved):

PR #1212, opened 2022-03-13
fix piece shine on firefox

nstringham:
this fixes a visual bug in browsers that support [contain](https://caniuse.com/mdn-css_properties_contain) but not [content-visibility](https://caniuse.com/css-content-visibility)

what it looks like when it's broken:
image

Chrome: already works
Firefox: works after this PR is merged
Safari: works after this PR is merged for users with the latest iOS/macOS

### DIFF
--- a/public/resources/scss/stats.scss
+++ b/public/resources/scss/stats.scss
@@ -1226,6 +1226,7 @@ a.additional-player-stat:hover {
   transition: box-shadow 0.2s ease-in-out;
   box-shadow: rgba(var(--text-rgb), 0) 0 0 0 0px inset;
   content-visibility: auto;
+  contain: paint;
 
   &.sticky-stats,
   &:focus {


dukio:
Requiring the latest (actually not even released yet) version of iOS to make the glint work.... like... is there no other way?
I understand if we do not support old browser/os... but we should at least support the most recent ones

nstringham:
> Requiring the latest (actually not even released yet) version of iOS to make the glint work.... like... is there no other way? I understand if we do not support old browser/os... but we should at least support the most recent ones

Yeah we could do some funny business with overflow but it doesn't seam worth it to me.

- most users don't have a Pioneer Pickaxe
- when it's broken it's not a very big deal
- it would hurt performance of other browsers (untested)
- we would have to come and change it latter
- do we even have a lot of iOS users?
  - you can't play skyblock on iOS
- we are going to want users on 15.4 anyways because the finally fixed scrolling

dukio:
> > Requiring the latest (actually not even released yet) version of iOS to make the glint work.... like... is there no other way? I understand if we do not support old browser/os... but we should at least support the most recent ones
> 
> Yeah we could do some funny business with overflow but it doesn't seam worth it to me.
> 
> * most users don't have a Pioneer Pickaxe
> * when it's broken it's not a very big deal
> * it would hurt performance of other browsers (untested)
> * we would have to come and change it latter
> * do we even have a lot of iOS users?
>   
>   * you can't play skyblock on iOS
> * we are going to want users on 15.4 anyways because the finally fixed scrolling

It's not just Pioneer Pickaxe but every mythic+ item (when piece-shine gets fixed)... it's kind of a big deal when it's a lot of items having that issue.

```css
.piece {
    ...
   overflow: hidden;
}
```

I don't get why we can't just put overflow hidden 😕

nstringham:
> It's not just Pioneer Pickaxe but every mythic+ item (when piece-shine gets fixed)... it's kind of a big deal when it's a lot of items having that issue.

what other mythic items are there?

> ```css
> .piece {
>     ...
>    overflow: hidden;
> }
> ```
> 
> I don't get why we can't just put overflow hidden 😕

You can try and see what it breaks but overflow hidden creates a stacking context which has performance implications as well as a bunch of other weird behavior.

dukio:
> You can try and see what it breaks but overflow hidden creates a stacking context which has performance implications as well as a bunch of other weird behavior.

Uhm if it has performance implications then let's just go with `contain: paint`... if we get users complaining about it then we can think of another solution (like overflow hidden) but for now this is fine then

dukio (approved):

nstringham:
Update: Safari 15.4 is now stable and users can upgrade whenever they want

MartinNemi03 (approved):

dukio:
Just noticed: this removes/hides the `select-weapon` checkbox below weapons
image

Same happens with `overflow: hidden`... needs another solution

nstringham:
@dukio thanks for catching that.

I hate that checkbox it causes so may problems because it is absolutely positioned outside the item

PR #1222, opened 2022-03-14
Essence section (moved outside of dungeons)

dukio:
Moved essence section outside of dungeons to prepare for the new nether update which will add the crimson essence, a new essence found in the nether and so not related to dungeons.

- Essence section is right below dungeons
- Missing essences (for which the user has 0 of) will be greyed out and semi-opaque (opacity 0.75)
- Essence object moved from `calculated.dungeon.essence` to `calculated.essence`

### Previews:
With some missing essences:
image

With all the essences:
image

Without any essence:
image


dukio:
Updated to include crimson essence (idk the real head for now so there's something similar as a placeholder)

metalcupcake5:
crimson essence head: `67c41930f8ff0f2b0430e169ae5f38e984df1244215705c6f173862844543e9d` will merge after

metalcupcake5:
crimson essence (real)

PR #1223, opened 2022-03-14
Icomoon unicode icons

### LINKED ISSUE #740: Use @font-face unicode-range feature with an svg-icon-font for symbols
Currently we are using an svg-icon-font for stars/master stars on items, .svg icons (applied with css image-mask) for player stats and normal characters rendered with the normal font in the lore.

Instead of doing all this we could simply maintain an svg-icon-font that customizes the aspect for some specific characters (stats symbols and stars/master stars) using the css @font-face unicode-range property.

How? 
- https://css-tricks.com/modifying-specific-letters-with-css-and-javascript/
- https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range (with multiple values syntax)

The support is also really good: [caniuse](https://caniuse.com/font-unicode-range)

dukio:
Closes #740 

Instead of using svg icon files we now have a custom font that replaces specific unicode characters with an svg-font from IcoMoon. This means that we can just type whatever skill symbol and it will be automatically replaced by the SVG Icon Font!

- Updated IcoMoon font to include all skills icons
- Updated IcoMoon `@font-face` to only replace specific UNICODE characters
- IcoMoon font is now applied to the whole website, on top on the normal Montserrat
- `display_name_print` is no longer needed so it has been deprecated and removed
- Updated README.md of IcoMoon font with instructions on how to add or edit icons for the future
- Fixed a bug where `bonus_attack_speed` was named `attack_speed`
- Fixes some pet perks lore where skill symbols didn't have a space between their value and the symbol

---
Current missing icons (but not really needed for now):
- True Damage
- Powder

# Preview
They are not quite as high quality as directly using SVGs, but they now apply on pet lore and all other lores in automatic, so I guess that balances it... 
Also you know... they don't look as good mainly on the stats section... and that might not be a problem anymore soon 👀 
## Before
image
image
image

## After
image
image
image


dukio:
Fixed conflicts (and idk why I had to remove some extra css which wasn't an issue before .-.)

public/resources/scss/shared.scss
```diff
@@ -883,15 +883,6 @@ input[type="search"]::-webkit-search-cancel-button {
...
 
-  &::before {
-    content: "";
-    display: inline-block;
-    vertical-align: middle;
-    height: 24px;
-    width: 24px;
-    background-color: currentColor;
-  }
-
```
nstringham, line 894: why are we removing this?

nstringham (commented):
I haven't tested it yet but you're code seams mostly fine

public/resources/scss/shared.scss
nstringham: oh wait I see what is going on here

dukio: Yeah it was the base background used for the svg mask

MartinNemi03 (approved):

PR #1232, opened 2022-03-18
Promotions

metalcupcake5:
- adds google form to the promotions
- allows "weighting" promotions to display some more than others
  - could probably be improved

image

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -512,7 +512,7 @@ async function processItems(base64, customTextures = false, packs, cacheOnly = f
       item.extra.skin = `PET_SKIN_${item.tag.ExtraAttributes.petInfo.skin}`;
     }
 
-    if(item.tag?.ExtraAttributes?.dye_item != undefined) {
+    if (item.tag?.ExtraAttributes?.dye_item != undefined) {
       item.extra.dye = item.tag.ExtraAttributes.dye_item;
     }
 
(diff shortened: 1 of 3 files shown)


src/lib.js
```diff
@@ -512,6 +512,10 @@ async function processItems(base64, customTextures = false, packs, cacheOnly = f
       item.extra.skin = `PET_SKIN_${item.tag.ExtraAttributes.petInfo.skin}`;
     }
 
+    if (item.tag?.ExtraAttributes?.dye_item != undefined) {
+      item.extra.dye = item.tag.ExtraAttributes.dye_item;
```
dukio: Ehm... wrong base branch?

dukio:
About the randomWeighted function did you write or find it?

I was looking at https://dev.to/trekhleb/weighted-random-algorithm-in-javascript-1pdc and their implementation is slightly different... if you found it I'm just wondering if it's accurate (the fact that it's not using `const` where it should be or returning from the last `for` instead of breaking out of it is kinda worring me)

metalcupcake5:
> did you write or find it

i found it, i do agree it could use some changing though. did some testing and it seems mostly accurate, i'll change a few things tho

dukio (changes requested):

dukio:
> > did you write or find it
> 
> i found it, i do agree it could use some changing though. did some testing and it seems mostly accurate, i'll change a few things tho

👍

src/lib.js
metalcupcake5: yikes mb

dukio (approved):

Issue #1218, opened 2022-03-14
Refactor/optimize `piece-shine` and `is-enchanted`


dukio:
(`piece-shine` is the item background animation, `is-enchanted` is the enchant glint applied as a mask on item icons)

---
Currently `piece-shine` and `is-enchanted` are a bit weird, this is how I would change it:
- Both properties are no longer calculated on the frontside but rather on the backend, at the end of `getItems()`
- `piece-shine` is renamed to `is-highlighted` (to prevent confusion with the new `shiny` item attribute)
- the item object will end up with 2 new properties: `isEnchanted` and `isHighlighted`, both booleans

`isEnchanted` will be true if:
- item has vanilla `ench` attribute (`Array.isArray(item.tag.ench)`)
- item has the new `shiny` property (making issue #1202 obsolete), should be automatically handled by the condition above
- item is one of [enchanted book, bottle o' enchanting, nether star]
- item is a potion with effects (so not just a water bottle)
- item is **_not_** a custom head

`isHighlighted` will be true if:
- item rarity is one of [very special, special, supreme, divine, mythic]

---
Currently there's also `item.glowing` property which was added to make some items in the HotM menu enchanted, like they are in game. This property can be replaced with `item.isEnchanted` directly so `glowing` will be deprecated.

---
A cool enhancement would be to make the glint effect more visible, maybe with some sort of animation... needs to be tested

nstringham:
how does the Minecraft client know whether or not to shoe a glint on shiny items?

if there is some nbt then we should look directly at that

other than that this seams like a good idea

dukio:
> how does the Minecraft client know whether or not to shoe a glint on shiny items?
> 
> if there is some nbt then we should look directly at that
> 
> other than that this seams like a good idea

I think it just checks for the `item.tag.ench` to exist, shiny items in /ah have `item.tag.ench = []` and are showing the glint while other items (not shiny) without `item.tag.ench` are not showing any glint.

Edited the issue so now it checks if the array exists at all, not if length > 0

DuckySoLucky:
@metalcupcake5

PR #1236, opened 2022-03-19
fix piece shine

dukio:
This PR replaces #1212 

- Removes `contain: paint` from `.piece` which was causing the issue of checkboxes not being shown
- Changes `.piece-shine` so that it's a container for the effect which is in the element `::before`
- Uses `@supports` for `contain: paint` so if the browser supports it we use it, otherwise it falls back to `overflow: hidden` (which is a little bit worse on performance side)
- Makes `.piece-shine` appear again on all mythic+ items (mythic, divine, supreme, special, very_special)... before it was bugged and only shown on divine+ (in #1218 will be refactored, but this is a first step towards it)
- Formatted the weapon section in `stats.ejs`... so I could actually read what was going on...

Tested on Firefox 98.0.1:
image


MartinNemi03 (approved):

nstringham:
@dukio sorry that I haven't had time to do skycrypt stuff. I would love to review this PR but don't wait for me because I don't know when the next time I will get free time is.

MartinNemi03:
> @dukio sorry that I haven't had time to do skycrypt stuff. I would love to review this PR but don't wait for me because I don't know when the next time I will get free time is.

No problem, thanks for letting us know, I'll go ahead and merge this then. Good luck with whatever you need to do!

PR #1263, opened 2022-03-30
Fix css stuff

### LINKED ISSUE #1257: Small css thing
**Is your suggestion related to a problem? Please describe in detail.**
A clear and concise description of what the problem is.
Hey, when using autofill on the search bar, it turns the background of the bar white 👎 Thats ugly ngl!

**Describe your suggestion**
Just override the default styling for :autocomple and :-webkit-autofill on the input

**Describe alternatives you've considered**
maybe making it autosearch after you autofill but that could be annoying if you misspelled it one time and that is now saved in your cache and when you click on it, it brings up the wrong player

**Additional context**
image
⬆️ What we currently have.
⬇️ What we could have!
image

### Now that ⬆️ is much better
(I typed that out btw to make it look good)

dukio:
Fixes #1257

- Removes an extra comment
- Removes duplicate css properties
- When duplicate css properties are there for a reason, wrap the second one inside `@supports`
- Change the cursor on clickable stuff to be a pointer
- Remove unit from 0 values (`0px` -> `0`)
- Adds a css hack to prevent browser (chrome/brave, maybe other chromium based browsers) from changing input background if it's from autofill with `transition: background-color 600000s 0s, color 600000s 0s;`

### DIFF
--- a/public/resources/scss/index.scss
+++ b/public/resources/scss/index.scss
@@ -65,13 +65,6 @@ main.grid {
       text-align: center;
       font-weight: 600;
       margin: 5px;
-
-      // //this is for random player
-      // &:last-child {
-      //     font-size: 16px;
-      //     color: var(--text-rgb);
-      //     margin: 0;
-      // }
     }
 
     input {
@@ -80,7 +73,6 @@ main.grid {
       height: 2.25em;
       text-align: center;
       font-size: inherit;
-      font-weight: 600;
       box-sizing: border-box;
       background-color: rgba(var(--text-rgb), 0.1);
       border: 0;
@@ -88,6 +80,11 @@ main.grid {
       font-weight: normal;
       -webkit-appearance: none;
       border-radius: 0;
+
+      &:-webkit-autofill,
+      &:-webkit-autofill:focus {
+        transition: background-color 600000s 0s, color 600000s 0s;
+      }
 
       &::-webkit-search-cancel-button {
         position: absolute;
@@ -109,7 +106,7 @@ main.grid {
       appearance: none;
       background-color: var(--icon-hex);
       color: white;
-      text-shadow: 0px 0px 3px rgba(0, 0, 0, 0.5);
+      text-shadow: 0 0 3px rgba(0, 0, 0, 0.5);
       text-transform: uppercase;
       font-weight: 700;
       font-size: 16px;
@@ -167,10 +164,12 @@ main.grid > a,
 #error_box_wrapper {
   position: relative;
   grid-column: 1 / -1;
-  overflow: hidden;
-  overflow: clip;
   min-height: 70px;
   max-height: 500px;
+  overflow: hidden;
+  @supports (overflow: clip) {
+    overflow: clip;
+  }
 
   &:not(.show-error) {
     display: none;
@@ -269,7 +268,9 @@ main.grid > a,
   height: 64px;
   border-radius: 5px;
   image-rendering: crisp-edges;
-  image-rendering: pixelated;
+  @supports (image-rendering: pixelated) {
+    image-rendering: pixelated;
+  }
 }
 
 .profile-name {
--- a/public/resources/scss/shared.scss
+++ b/public/resources/scss/shared.scss
@@ -607,7 +607,7 @@ input[type="search"]::-webkit-search-cancel-button {
 }
 
 .expander {
-  cursor: context-menu;
+  cursor: pointer;
 
   &::before {
     content: "";
@@ -1226,4 +1226,9 @@ body {
     line-height: 22px;
     height: 22px;
   }
+}
+
+button[name="pack"],
+input[name="theme"] {
+  cursor: pointer;
 }
--- a/public/resources/scss/shared.scss
+++ b/public/resources/scss/shared.scss
@@ -308,7 +308,9 @@ header {
   z-index: 1000;
   line-height: 48px;
   overflow: hidden;
-  overflow: clip;
+  @supports (overflow: clip) {
+    overflow: clip;
+  }
 }
 
 header > * {
@@ -452,6 +454,11 @@ header .lookup-player {
       outline: none;
       background-color: rgba(0, 0, 0, 0.2);
     }
+
+    &:-webkit-autofill,
+    &:-webkit-autofill:focus {
+      transition: background-color 600000s 0s, color 600000s 0s;
+    }
   }
 
   button {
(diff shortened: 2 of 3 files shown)


nstringham:
> When duplicate css properties are there for a reason, wrap the second one inside @supports

why?

nstringham:
> Change the cursor on clickable stuff to be a pointer

there is some debate on this but it looks like this is not recommended
this article seams to do a good job of explaining it https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b

nstringham:
> Remove unit from 0 values (0px -> 0)

this seams like it reduces readability because it makes it look like a unitless value whereas the unit makes it clear what the unit is

basically `0px` is obviously not `0deg` or `0s` but `0` could be anything

dukio:
> > Remove unit from 0 values (0px -> 0)
> 
> this seams like it reduces readability because it makes it look like a unitless value whereas the unit makes it clear what the unit is
> 
> basically `0px` is obviously not `0deg` or `0s` but `0` could be anything

0 is 0... no matter the unit (px, rem, em, ch, vh, vw, ...), and it's only removed on length measures (width/height, margin, ...)

dukio:
> > Change the cursor on clickable stuff to be a pointer
> 
> there is some debate on this but it looks like this is not recommended this article seams to do a good job of explaining it https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b

the article makes a good point but I also agree with some comments in that article... 
like that `pointer: cursor` just adds a tiny feedback extra for the user that the button is clickable, and will do something
also we don't have any other `:hover` animation, so the cursor at least is something
and that all the buttons having similar style should work the same way, so have the same hover effect and cursor, currently it's not like that:
image

Red things have `cursor: pointer`, the rest have `cursor: default`... so if the user goes over API and sees a pointer cursor it might think ABOUT doesnt do anything, cause it doesn't have one... similar on the right with Themes and Packs or on the left with the logo (even though that's more of an easter egg)

So... if it makes it easier for the user to understand that the button is clickable, even by a tiny bit, I think we should add it... doesn't hurt anyone

dukio:
> > When duplicate css properties are there for a reason, wrap the second one inside @supports
> 
> why?

1. So VSCode doesn't spam warnings "duplicate property on line xxx"
2. If we have duplicate properties, is to give modern browsers something more with a fallback for ancient ones... and that's the purpose of `@supports`, so I think it makes everything more "semantic"
3. In the future, when a property that is now inside a `@support` cause of low browser supports becomes available to a lot of users thanks to browser updates we can easily spot and update them

nstringham:
> 0 is 0... no matter the unit (px, rem, em, ch, vh, vw, ...), and it's only removed on length measures (width/height, margin, ...)

it still seams more obvious if it has a unit but I guess if `0` is always distance then it is ok

nstringham:
> > > Change the cursor on clickable stuff to be a pointer
> > 
> > 
> > there is some debate on this but it looks like this is not recommended this article seams to do a good job of explaining it https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b
> 
> the article makes a good point but I also agree with some comments in that article... like that `pointer: cursor` just adds a tiny feedback extra for the user that the button is clickable, and will do something also we don't have any other `:hover` animation, so the cursor at least is something and that all the buttons having similar style should work the same way, so have the same hover effect and cursor, currently it's not like that: image
> 
> Red things have `cursor: pointer`, the rest have `cursor: default`... so if the user goes over API and sees a pointer cursor it might think ABOUT doesnt do anything, cause it doesn't have one... similar on the right with Themes and Packs or on the left with the logo (even though that's more of an easter egg)
> 
> So... if it makes it easier for the user to understand that the button is clickable, even by a tiny bit, I think we should add it... doesn't hurt anyone

that's true but I think we should make it clear they are interactable in other ways. especially because a lot of users don't have pointers. for example keyboard navigation and touch screens. We could have a `:hover` and `:focus` for keyboard and mouse support I think that for mobile users mostly use the way things look to decide what is clickable.

nstringham:
> 1. So VSCode doesn't spam warnings "duplicate property on line xxx"

what extension are you using? I don't have that

> 2. If we have duplicate properties, is to give modern browsers something more with a fallback for ancient ones... and that's the purpose of `@supports`, so I think it makes everything more "semantic"

I think `@supports` is for more complex situations and adding a new rule seams a bit excessive.

> 3. In the future, when a property that is now inside a `@support` cause of low browser supports becomes available to a lot of users thanks to browser updates we can easily spot and update them

what if we just put a comment like
```scss
&.pack-icon {
    image-rendering: crisp-edges; // fallback for pixelated
    image-rendering: pixelated;
}
```

dukio:
> > 1. So VSCode doesn't spam warnings "duplicate property on line xxx"
> 
> what extension are you using? I don't have that
> 
> > 2. If we have duplicate properties, is to give modern browsers something more with a fallback for ancient ones... and that's the purpose of `@supports`, so I think it makes everything more "semantic"
> 
> I think `@supports` is for more complex situations and adding a new rule seams a bit excessive.
> 
> > 3. In the future, when a property that is now inside a `@support` cause of low browser supports becomes available to a lot of users thanks to browser updates we can easily spot and update them
> 
> what if we just put a comment like
> 
> ```css-scss
> &.pack-icon {
>     image-rendering: crisp-edges; // fallback for pixelated
>     image-rendering: pixelated;
> }
> ```

No clue honestly, I think it's builtin
image

image

dukio:
> > > > Change the cursor on clickable stuff to be a pointer
> > > 
> > > 
> > > there is some debate on this but it looks like this is not recommended this article seams to do a good job of explaining it https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b
> > 
> > 
> > the article makes a good point but I also agree with some comments in that article... like that `pointer: cursor` just adds a tiny feedback extra for the user that the button is clickable, and will do something also we don't have any other `:hover` animation, so the cursor at least is something and that all the buttons having similar style should work the same way, so have the same hover effect and cursor, currently it's not like that: image
> > Red things have `cursor: pointer`, the rest have `cursor: default`... so if the user goes over API and sees a pointer cursor it might think ABOUT doesnt do anything, cause it doesn't have one... similar on the right with Themes and Packs or on the left with the logo (even though that's more of an easter egg)
> > So... if it makes it easier for the user to understand that the button is clickable, even by a tiny bit, I think we should add it... doesn't hurt anyone
> 
> that's true but I think we should make it clear they are interactable in other ways. especially because a lot of users don't have pointers. for example keyboard navigation and touch screens. We could have a `:hover` and `:focus` for keyboard and mouse support I think that for mobile users mostly use the way things look to decide what is clickable.

I agree... it can for sure be improved with hover and focus states, but this is something to start width

MartinNemi03 (approved):

metalcupcake5 (approved):

PR #1269, opened 2022-04-06
Removing unused constants and typo fix

Korase:
I was looking through the constants files and removed unused ranks in misc, the survey we finished a decent bit ago, as well as fixing a single typo in a comment I read.

### DIFF
--- a/src/constants/misc.js
+++ b/src/constants/misc.js
@@ -248,6 +248,16 @@ export const ranks = {
     plusColor: "b",
   },
 
+  MAYOR: {
+    color: "d",
+    tag: "MAYOR",
+  },
+
+  MINISTER: {
+    color: "c",
+    tag: "MINISTER",
+  },
+
   NONE: null,
 };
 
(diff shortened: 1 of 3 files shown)


dukio:
About the ranks, ok on build team cause I don't think it will come back, but for mayor and minister we can't be sure it won't come back so I'd leave it there

For the survey, it's a new one we made a couple weeks ago I think, maybe it's been a bit more but it's not that long ago

For the comment woooops my bad

Korase:
> About the ranks, ok on build team cause I don't think it will come back, but for mayor and minister we can't be sure it won't come back so I'd leave it there
> 
> For the survey, it's a new one we made a couple weeks ago I think, maybe it's been a bit more but it's not that long ago
> 
> For the comment woooops my bad

I was under the impression we had already finished the survey as we had discussed the results a while ago and the results aren't subject to change much further.

nstringham (approved):

MartinNemi03 (approved):

dukio (approved):

dukio:
Ok yeah we got 1.1k answers and it's probably best to close the survey before nether update drops

PR #1271, opened 2022-04-07
Remove sass `:where()` workaround

nstringham:
we no longer need this workaround because it was fixed in https://github.com/sass/sass/pull/3283


dukio (approved):
lgtm didn't test

Issue #1268, opened 2022-04-06
Missing period in pet tooltips


DespicableGoose:
**Missing period in pet tooltips**
At the end of pet abilities in-game, there is generally a period. However, this period is omitted on SkyCrypt. I am not sure whether or not this applies to other tooltips as well.

**Profile/Location**
Any, although mine (TheTrueShaman) and Refraction were those used to test this (refraction because he has lots of pets).

**Expected behavior**
There would be a period at the end of the line such as "Zaps your enemies for 20x your ✎ Intelligence every 3s."

**Screenshots**
In-game:
<img width="361" alt="Screen Shot 2022-04-05 at 10 56 56 PM" src="https://user-images.githubusercontent.com/68354378/161912622-09470336-7278-4d52-8e03-37e535efc03d.png">
SkyCrypt:
<img width="350" alt="Screen Shot 2022-04-05 at 11 49 08 PM" src="https://user-images.githubusercontent.com/68354378/161912713-4a02dacd-8543-4eac-9520-d41ca211d1fc.png">

**Desktop settings:**
- OS: MacOS Monterey 12.3
- Browser: Google Chrome
- Version 100.0.4896.75 (Official Build) (x86_64)

dukio:
Peiords in pet perks are random... on some perks there final dots, in others it's missing... doesn't really matter tbh, usually we don't add the final periods for new pets's perks, there might be some pets on skycrypt who have the final dot

Harteiga:
I looked into it and it's not that this was a deliberate choice but that it was originally the way skycrypt displays it. see: https://youtu.be/1SOyNv6SJbw?t=100. This was later changed at some unknown date. That might be why it looks random.

Though this is a little off topic. I noticed the Grandma Wolf Pet was changed.

New:
![image](https://user-images.githubusercontent.com/100935552/162091852-94541020-017e-4538-8254-8dc4234245da.png)
Current Displayed:
![image](https://user-images.githubusercontent.com/100935552/162092234-62568846-96aa-4b60-90d9-b0ee2843b71d.png)

PR #1285, opened 2022-04-11
fix grandma wolf perk description

dukio:
As reported by #1268 (but that issue is not fixed yet), fix the description of grandma wolf pet:

New:
image

Before:
image


dukio:
In game: 
image

MartinNemi03 (approved):

PR #1293, opened 2022-04-17
Cache subfolders

MartinNemi03:
Added type subfolders for /cache folder with some folder optimization in mind.
This is not applied in development mode, so it won't spam you with folders. 
It's mainly meant for the main site, where we can have tens of thousands of files in a single folder.

### DIFF
--- a/src/helper.js
+++ b/src/helper.js
@@ -993,9 +993,14 @@ export function getCacheFilePath(dirPath, type, name) {
 
   let subdirs = [type];
 
-  // for texture and head type, we want it to be a bit more split between folders
+  // for texture and head type, we get the first 2 characters to split them further
   if (type == "texture" || type == "head") {
     subdirs.push(name.slice(0, 2));
+  }
+
+  // for potion and leather type, we get what variant they are to split them further
+  if (type == "leather" || type == "potion") {
+    subdirs.push(name.split("_")[0]);
   }
 
   // check if the entire folder path is available
--- a/src/helper.js
+++ b/src/helper.js
@@ -991,7 +991,7 @@ export function getCacheFilePath(dirPath, type, name) {
     return path.resolve(dirPath, `${type}_${name}.png`);
   }
 
-  let subdirs = [type];
+  const subdirs = [type];
 
   // for texture and head type, we get the first 2 characters to split them further
   if (type == "texture" || type == "head") {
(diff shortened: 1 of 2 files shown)


src/helper.js
```diff
@@ -983,6 +985,40 @@ export function parseItemTypeFromLore(lore) {
   };
 }
 
+export function getCacheFilePath(dirPath, type, name) {
+  // we don't care about folder optimalization when we're developing
```
dukio: optimization typo

src/helper.js
```diff
@@ -983,6 +985,40 @@ export function parseItemTypeFromLore(lore) {
...
+  if (process.env?.NODE_ENV == "development") {
+    return path.resolve(dirPath, `${type}_${name}.png`);
+  }
+
+  let subdirs = [type];
```
dukio: const?

src/helper.js
```diff
@@ -983,6 +985,40 @@ export function parseItemTypeFromLore(lore) {
...
+  if (type == "leather" || type == "potion") {
+    subdirs.push(name.split("_")[0]);
+  }
+
+  // check if the entire folder path is available
+  if (!fs.pathExistsSync(path.resolve(dirPath, subdirs.join("/")))) {
+    // check if every subdirectory is available
+    for (let i = 1; i <= subdirs.length; i++) {
+      let checkDirs = subdirs.slice(0, i);
+      let checkPath = path.resolve(dirPath, checkDirs.join("/"));
```
dukio: all const?

dukio (changes requested):
I'm on mobile but I noticed these things... will check/test further when on desktop

src/helper.js
MartinNemi03: not sure, since i potentially push into it, but i'll try.

src/helper.js
MartinNemi03: ty, my english broken.

src/helper.js
MartinNemi03: yes, all const.

src/helper.js
dukio: Yeah pushing is not an issue, only thing you can't do is reassign the value completely

dukio (approved):
didn't test but looks good

metalcupcake5 (approved):

PR #1306, opened 2022-04-20
Frontend stats

dukio:
This is the first of 2 PRs (second one will be created at a later date, probably once all the new nether update stuff has been handled).
1. PR that replaces stats in the frontend with frontend calculated stats, in the backend they are still calculated but not displayed anymore (this PR)
2. Second PR will clean up everything in backend that calculates stats that is no longer needed (should be almost everything)

Pretty much, merge this PR and you should see no change at all... yay 🥳 
No but for real... old backend stats and new frontend stats are really close to each-other, and in some instances frontend is even more accurate.

Honestly this PR does a lot for the handling of stats which is hard to write down here... but I'll try:
- common/constants can now be splitted in multiple files for better code readability
- new constants `statsData` object, which includes all the commonly needed things from stats (name, short name, color, symbol, etc...)
- calculates player stats in frontend
- displays stats using new lit component `player-stat`ù
- displays bonus stats from certain things (like armor, accessories, pet, slayer) using new lit component `bonus-stats`
- cleans up some old constants and moves some to the common folder to be shared by both.
- backend `helper` and `constants` now grab stuff from `common` folder too
- all this also ended up cleaning `stats.ejs` from some hardcoded constants there and a couple functions replaced by the lit components


metalcupcake5 (approved):
didnt test

MartinNemi03 (approved):
other than selecting items being gone (hope that comes back at a later date), lgtm!

PR #1343, opened 2022-04-29
Fix Odger's tooth order

nstringham:
before this PR the progression was bronze -> gold -> diamond -> silver
after this PR the progression goes bronze -> silver -> gold -> diamond

profile to use to verify this change https://cupcake.shiiyu.moe/stats/HsFearless/Cucumber#Accessories

original bug report
image


MartinNemi03 (approved):

dukio (approved):
lgtm didn't test

PR #1345, opened 2022-04-30
Fix JS failing when player hasn't done catacombs

Korase:


dukio (approved):
looks good to me

But do you have a link to a profile where this failed? 🤔 
I tested with https://sky.shiiyu.moe/stats/DynamicEx/Lemon#Dungeons (a player who never played dungeons) and the JS is not failing 😕

dukio (changes requested):
Actually hold on... maybe we should just check if the `calculated.dungeons.catacombs.level` exists, otherwise skip the entire block

Yeah I think that's cleaner... so just wrap everything in

```js
if (calculated.dungeons?.catacombs?.level) {
 // ...
}
```

dukio:
I went ahead and proposed a fix with Korase#1

dukio (approved):

MartinNemi03 (approved):

PR #1350, opened 2022-05-01
Changes to the system for featured profiles 

### LINKED ISSUE #1348: Featured profiles don't account for name changes
When a featured profile changes their name ingame, they no longer show up on the frontpage.

MartinNemi03:
- Changed featured profiles to use UUID instead of usernames
- Changed the system to export into a JSON file instead of a database

Fixes #1348

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -140,6 +140,8 @@ function updateCommitHash() {
 }
 const commitHash = updateCommitHash();
 
+const featuredProfiles = await fs.readJSONSync(path.resolve("./public/resources/js/featured-profiles.json"));
+
 const app = express();
 const port = process.env.SKYCRYPT_PORT ?? 32464;
 
@@ -223,8 +225,7 @@ async function getExtra(page = null, favoriteUUIDs = [], cacheOnly) {
 
   if (page === "index") {
     output.favorites = await getFavoritesFormUUIDs(favoriteUUIDs);
-
-    output.devs = await db.collection("featuredProfiles").find().sort({ position: 1 }).toArray();
+    output.featured = featuredProfiles;
   }
 
   return output;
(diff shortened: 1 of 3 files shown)


src/app.js
```diff
@@ -140,6 +140,8 @@ function updateCommitHash() {
 }
 const commitHash = updateCommitHash();
 
+const featuredProfiles = await fs.readJSONSync(path.resolve("./public/resources/js/featured-profiles.json"));
```
nstringham: why is there an `await` here?

dukio:
But why is it needed to update them? nothing gets updated... it's just a static object no? 😕 🤔 

The whole `/scripts/update-featured-profiles.js)` seems useless to me, no?

nstringham:
What about skins are those still going to auto update?

dukio:
> What about skins are those still going to auto update?

Yeah they are from crafatar, so external

nstringham:
> > What about skins are those still going to auto update?
> 
> Yeah they are from crafatar, so external

Good point

MartinNemi03:
> But why is it needed to update them? nothing gets updated... it's just a static object no? 😕 🤔
> 
> The whole `/scripts/update-featured-profiles.js)` seems useless to me, no?

You need to update the usernames and emojis somehow. 
Also that's why it only runs on start, it only needs to get updated user info and that's it.

dukio:
> > But why is it needed to update them? nothing gets updated... it's just a static object no? 😕 🤔
> > The whole `/scripts/update-featured-profiles.js)` seems useless to me, no?
> 
> You need to update the usernames and emojis somehow. Also that's why it only runs on start, it only needs to get updated user info and that's it.

oh right didn't think of that

dukio:
#1358

src/app.js
dukio: Agree it doesn't look like the await is needed: https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/readJson-sync.md

dukio (approved):
👍

PR #1380, opened 2022-05-11
Change let to const when not re-assigned

Korase:
After doing this for pet-stats constants for dukio, I decided it would be a good fit to do it for every single file.


dukio:
Will check this in a bit but... do you have a script or something to find these? or did you go through all fines manually 😅 ?

Korase:
Manually 😅

dukio (approved):
quickly looked through the changes in github but also tested some profiles in localhost and it works fine so 👍

MartinNemi03 (approved):
LGTM!

PR #1390, opened 2022-05-12
API routing

MartinNemi03:
API now has routes, because it was really a mess. Now it's hopefully less of a mess.
Both API and APIv2 should be working as they did before with better handling of errors and unknown endpoints.

_Also now I know why Lea abandoned the previous API and just made V2._


dukio (approved):
seems fine to me... honestly I didn't check API much cause I never used it before but from code it looks good and it starts in local so... gtm

PR #1437, opened 2022-05-29
Fix pets TB & Wisp Pet

### LINKED ISSUE #1429: Frost Wisp & upgrades
Wisp pet changes ID every tier 😑 

...handle it... somehow...

also make sure it only counts once on the "total unique pets"

Pensul has it in /pets

### LINKED ISSUE #1430: Tierboost on pets
Verify on Griffin and Kuudra:

https://sky.shiiyu.moe/stats/DeathStreeks/Blueberry

On Wisp it doesn't work, but that's because each tier has a different ID, so stats are correlated to that ID

dukio:
Fixes #1430
Fixes #1429

So... for some reason Wisp is the only pet which changes it's TYPE based on the tier... it's upgraded through "upgrade stones" like Griffin but this one had to change TYPE too I guess 😑 

So this PR basically handles that...

Please note... a TierBoost Droplet Wisp (uncommon) in game shows like a rare one, but to keep things not too complicated on SkyCrypt it will still display as uncommon. Why?
- To support multiple TB at this point we would have to implement duplicate code, which is not good
- TB can't be applied to Wisp pet anymore, so only a few people have them
- Even if someone has TB Wisp pet, the only difference between us and in-game is the color; the perks, name and texture is correct... just the color is not correct...

Griffin also doesn't allow TierBoost, so even though noone should be able to own a TB Griffin, I made sure it will display with the correct tier/perks anyway (it came free just by adding the option in the petData object).

This PR also fixes the display of Unique Pets and Total Pets...
- Now they both ignore Bingo Pet on non-bingo profiles
- Now they both support Wisp having multiple TYPEs (DROPLET_WISP, FROST_WISP, ...), to handle this you can now set the petData property "typeGroup", now it's only used for WISP.


dukio:
Frost Wisp can be tested on Pensul profile (only player with an upgraded Wisp afaik (admin))

MartinNemi03 (approved):

Issue #1393, opened 2022-05-13
Enhance stats


dukio:
This issues serves the purpose to track all things that could be improved or implemented to make stats more accurate.

## This is an ever evolving list

- [ ] Pet perks that alter weapon, armor or stats directly
- [ ] Magic find from pet score
- [ ] Century cakes, see #1386
- [ ] Potion effects, see #1382 
- [ ] Dungeon shop upgrades, see #1385 
- [ ] Harp quest, see #1384 
- [ ] Accessory bag power and tuning, see #1381 
- [ ] Bestiary level (+2hp per level)
- [x] Reaper peppers (+1hp each)
- [ ] Harp intelligence bonus
- Reforges:
  - [ ] Renown
  - [ ] Loving, see #1419
- Armor set bonuses or abilities
  - [ ] Superior Dragon Armor
  - [ ] Young Dragon Armor
  - [ ] Cheap Tuxedo
  - [ ] Fancy Tuxedo
  - [ ] Elegant Tuxedo
  - [ ] Lapis Armor
  - [ ] Emerald Armor
  - [ ] Fairy Armor
  - [ ] Speedster Armor
  - [ ] Mastiff Armor
- Item abilities
  - [ ] Day Crystal
  - [ ] Night Crystal
  - [ ] Obsidian Chestplate
  - [ ] Speed Talisman
  - [ ] Speed Ring
  - [ ] Speed Artifact
  - [ ] Frozen Chicken
- Items giving personalized bonus based on kills or other things ("Kills: ", "Your Bonus: ", ...)

EHP?

---
**Also see `lib.js->getStats()`**

DuckySoLucky:
~~Hey, letting u know that this stuff has been already done in~~ ~~[QwackAPI](https://github.com/DuckySoLucky/QwackAPI/blob/main/stats/stats.js).~~ ~~Accuracy isn't 100% there are some stuff which are not possible to make like Magic Find and Wisdom from Booster Cookie (Impossible to detect if player has an cookie or not, not in API I think, not 100% sure) and 1-10 Intelligence (Defuse Kit, not in API). Example of stats can be found~~ ~~[Here](https://github.com/DuckySoLucky/QwackAPI/blob/main/examples/v2/stats.json)~~.~~I have tested this on multiple profiles and accounts, accuracy is like 99% without taking in fact impossible stuff, mostly because Hypixel's WIKI's reforge information isn't 100% accurate.~~

[...]

PR #1457, opened 2022-06-08
cake buffs

### LINKED ISSUE #1386: API temp_stat_buffs (century cakes, maybe other stuff too?)
image

metalcupcake5:
adds stats from century cakes
before:
image
after:
image

bonus listed in tooltip
image
image


metalcupcake5:
metal forgets to lint part 819289298981

src/lib.js
```diff
@@ -2516,6 +2516,22 @@ export const getStats = async (
     lily: calculateLilyWeight(output),
   };
 
+  /*
+
+    century cake effects
+
+  */
+  const cakes = [];
+  //console.log(userProfile.temp_stat_buffs);
```
dukio: remove extra console log

src/lib.js
```diff
@@ -2516,6 +2516,22 @@ export const getStats = async (
...
+  for (const cake of userProfile.temp_stat_buffs) {
+    const stat = cake.key.substring(cake.key.indexOf("_") + 1, cake.key.length);
```
dukio: why not just `const stat = cake.key.replace('cake_', '')` ?

also needs to add an `if (!cake.key.startsWith('cake_')) { continue }` in case in the future century cakes are not the only things inside this temp_stat_buffs

src/lib.js
```diff
@@ -2516,6 +2516,22 @@ export const getStats = async (
...
+    cakes.push({
+      stat: stat == "walk_speed" ? "speed" : stat,
+      amount: cake.amount,
+    });
+  }
+  output.cakes = cakes;
```
dukio: I feel like "cakes" is too generic... no?
I would use "century_cakes"

public/resources/ts/calculate-player-stats.ts
```diff
@@ -127,6 +127,16 @@ export function getPlayerStats() {
     }
   }
 
+  {
```
dukio: if everything is inside an `if` you dont need the outer `{}` scope, the `if` itself is the scope

public/resources/ts/calculate-player-stats.ts
```diff
@@ -127,6 +127,16 @@ export function getPlayerStats() {
...
+    if (calculated.cakes) {
+      for (const cake of calculated.cakes) {
+        if (stats[cake.stat]) {
+          stats[cake.stat].cakes ??= 0;
```
dukio: I would call them "century_cakes" here too so players can't confuse them with the cake bag

dukio (changes requested):
Just a few things to review but overall looks good 👍

dukio:
Would be cool to also display these things.... but that's gonna be a lot more work :/

It could go in an "effects" section

After this is merged close or change the scope of issue #1386 and tick the checkbox in #1393

dukio:
From the code looks good, I'll test it later today when I get back home

I'm also wondering if we should already code a mapping between potion effect => stat

Like for walk_speed = speed, there might be other cases (maybe not for century cakes but for all other potions maybe yes)

metalcupcake5:
i was planning on adding potion effects so if its the same mapping i'll do that then

dukio:
> i was planning on adding potion effects so if its the same mapping i'll do that then

If you want you can just do a mapping for walk speed so this pr can be merged

And when the potion effects pr gets done the missing mappings will be added

dukio (approved):

metalcupcake5:
metal forgets to lint part 819289298982

MartinNemi03 (approved):

PR #1488, opened 2022-06-11
fix fortunate

dukio:
Fixes FurfSkyReborn having wrong texture for maxed fortunate hotm


MartinNemi03 (approved):
seems like more of a bandaid for when we properly add things like these, but lgtm

dukio:
> seems like more of a bandaid for when we properly add things like these, but lgtm

image

This is the generated lore... guess I could add an extra `§6` before the `Fortune` but... feels worse than a simple comment

metalcupcake5 (approved):

PR #1491, opened 2022-06-11
Backend stats and lib.js cleanup

dukio:
In this PR I'm cleaning up `lib.js` (mostly) by removing all backend stats calculations (unless needed there), and other little things I notice along the way (that still aren't too far away from this PR scope)

- Removing stats calculations in backend
- Removing unused functions in lib.js and stats.ejs
- Removing stats from the profile pages description (og tag, the preview in discord... we don't have those anymore)
- Removing the export from functions not used outside of lib.js


dukio:
Ready for review.

I would still like to refactor `getItems()` and `getStats()`, but those probably need to be separate PRs, otherwise I'm blocking everyone else from development for way too long 😅

MartinNemi03 (approved):

metalcupcake5 (approved):

PR #1524, opened 2022-06-27
fix stats for french users

dukio:
Fix an issue with french language causing an error while rendering stats, more details here
https://discord.com/channels/738971489411399761/738990246825427084/991094407145066557


dukio:
Closing this, #1526 is a better solution

PR #1538, opened 2022-07-06
remove fallback for backdrop-filter

nstringham:
Firefox is about to finally add support for `backdrop-filter` in Firefox 103 (see [caniuse/css-backdrop-filter](https://caniuse.com/css-backdrop-filter))

once this feature is fully rolled out we can safely remove support for browsers that do not have this feature


nstringham:
Firefox 103 came out today

nstringham:
I'm not sure how long it will take for people to update

dukio:
From caniuse relative usage it looks like a lot of people are still using older versions....

metalcupcake5 (approved):

MartinNemi03:
I think we can safely assume now most Firefox users use 103.

PR #1541, opened 2022-07-07
add canonical links

nstringham:
this PR adds a canonical link to every page this has a few benefits when it comes to SEO:

I noticed that some links when searching for "SkyCrypt" on Google lead to players that have changed their names. by providing a canonical URL using a UUID this would continue to work even after a username change.

It will help prevent Google from linking to the HTTP version of the site (although this doesn't seem to be a problem for us)

It will prevent sites that run an unedited version of SkyCrypts code (such as cupcake.shiyyu.moe) from appearing in Google search results.

It will prevent duplicates from showing up in search results
  - example of a search with duplicates: https://www.google.com/search?q=SkyCrypt+Cookie_Wookie_7


MartinNemi03 (approved):
didn't test, lgtm!

dukio (commented):
Wouldn't you add a final else clause which just keeps the current path?

Just in case we add more pages in the future so they don't automatically get canonical to the home page

nstringham:
> Wouldn't you add a final else clause which just keeps the current path?
> 
> Just in case we add more pages in the future so they don't automatically get canonical to the home page

I thought of that but I couldn't think of a way to get the current path and I doubt we will add more very often. One option would be to put an else case that throws an exception

dukio:
> > Wouldn't you add a final else clause which just keeps the current path?
> > Just in case we add more pages in the future so they don't automatically get canonical to the home page
> 
> I thought of that but I couldn't think of a way to get the current path and I doubt we will add more very often. One option would be to put an else case that throws an exception

uhm... the exception could work, at least as a reminder for us

or just don't print the canonical link at all?

nstringham:
> or just don't print the canonical link at all?

That could work but it would allow sites like cupcake to show up in search results

dukio:
> > or just don't print the canonical link at all?
> 
> That could work but it would allow sites like cupcake to show up in search results

we have no way to detect if it's sky or cupcake right? otherwise a robots noindex could work to fix that issue

dukio (approved):
lgtm didn't test

Issue #1346, opened 2022-05-01
Update code that gets armor name


dukio:
[src/lib.js#L1348](https://github.com/SkyCryptWebsite/SkyCrypt/blob/92b031c935279627adfb9ed59931395a9fd17ba8/src/lib.js#L1348-L1352)
![image](https://user-images.githubusercontent.com/2744227/166144170-a5c9fe11-a624-4f5b-9253-9e19164eeb4d.png)

This should be updated to include the new dungeon "stars" things, armor dyes, dungeon boss enhancement (this thing: ![image](https://user-images.githubusercontent.com/2744227/166144096-dc860a26-e41e-4176-9dab-3725084a0873.png))

Or better at this point just do a whitelist regex for `A-Za-z0-9 -'`, and `.trim()`

StickyRunnerTR:
instead of regex A-Za-z0-9 -', use \w, far easier imo

dukio:
> instead of regex A-Za-z0-9 -', use \w, far easier imo

`\w` is equivalent to `A-Za-z0-9_`, not `A-Za-z0-9 -`

DuckySoLucky:
@metalcupcake5

PR #1574, opened 2022-08-06
Updated the code that parses the name of the active armor piece

lochyj:
Fixes issue:
https://github.com/SkyCryptWebsite/SkyCrypt/issues/1346

### DIFF
--- a/common/constants/enchantments.js
+++ b/common/constants/enchantments.js
@@ -31,7 +31,7 @@ export const maxEnchants = new Set([
   "Fire Aspect III",
   "Fire Protection VII",
   "First Strike V",
-  "Flame I",
+  "Flame II",
   "Fortune IV",
   "Frail VI",
   "Frost Walker II",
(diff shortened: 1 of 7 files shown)


nstringham (commented):
What even is `arnor_name`?

lochyj:
> What even is `arnor_name`?

Armor_name is the name of the armor set that is displayed above the active armor set, the old code removed the stars and the skin icon `✪⍟✦`. The new code removes all characters other than `a - z`, `A - Z`, `0 - 9` and `-`. which future proof it so if an admin decides to add another symbol to an armor piece, the new code will remove it. The code also removes the upgrade icon which addresses issue: https://github.com/SkyCryptWebsite/SkyCrypt/issues/1346
This fix was suggested in that issue

The following images show what the code will do
Before:
image
After:
image

dukio (approved):
lgtm, didn't test

dukio:
I checked the wiki pages that list all armor sets and `-` doesn't seem to be used in armor names currently... but still it's good to keep for future proofing (I can totally see an armor name having a - in the future)

nstringham:
@lochyj  you should use a feature branch. This PR is all messy now because it included other changes.

lochyj:
Now at:
https://github.com/SkyCryptWebsite/SkyCrypt/pull/1589
bc i messed up

PR #1635, opened 2022-09-03
Fix for "CONSTANTS is not defined"

MartinNemi03:
Basically there was one reference to uppercase `CONSTANTS` instead of lowercase `constants` which broke the rendering of `stats.ejs`.


metalcupcake5 (approved):

dukio (approved):
lgtm didn't test

PR #1639, opened 2022-09-05
Stats

### LINKED ISSUE #1603: 0.14 - New stats / stats changes
https://hypixel.net/threads/skyblock-patch-0-14-shens-auctions-the-hex-rng-meter-and-more.5074247/

> The stats menu(s) available in "your profile menu" has been upgraded to show where all of your stats are from.
> If you want to learn more about it, [here's an objectively mediocre guide about it](https://www.youtube.com/watch?v=KgcY8bNsoKk).
> New Stats:
7 Wisdom Stats for different skills, which increase skill XP gain. (For example, Slayer now grants Combat Wisdom!)
Fishing Speed
Health Regen
Vitality, which boosts incoming healing (good for tanks!)
Mending, which boosts outgoing healing (good for healers!)
Fishing Speed also works differently as before. It used to remove a % of the remaining time available to spawn the fish, the system now simply adds a value and it linearly makes the particles spawn faster.
If you reach 300 Fishing Speed, the particles will appear instantly, but there is an exception with the Crimson Isle, where 350 Fishing Speed is the value you have to reach for maximum efficiency.

metalcupcake5:
image
wip
closes #1603


public/resources/scss/stats.scss
```diff
@@ -899,46 +905,32 @@ a.additional-player-stat:hover {
...
+.color-intelligence
+.color-magic-find,
+.color-fishing-speed {
   color: var(--§b);
 }
 
 .stat-sea-creature-chance,
-.color-sea-creature-chance {
+.color-sea-creature-chance,
+[class $= "wisdom"] {
```
nstringham, line 918: What does this compile to?

dukio: Everything with a class that ends in wisdom (not sure about the spaces though).

@metalcupcake5 I think it's much much safer to specify each type of wisdom, I would do that

nstringham: Oh I thought this was a fancy sass thing I didn't realize that was built in CSS.

One issue is that this actually only selects things where the last class ends in wisdom

I think would be better to select based on actual classes instead of based on the class attribute

metalcupcake5: got it

PR #1650, opened 2022-09-11
feat: proper recomb accessory count

### LINKED ISSUE #1637: Pandora's Box shouldn't count towards recombobulated accessories
Since it can't be recombobulated but gets its rarity from the challenges completed by the player

metalcupcake5:
closes #1637 
image
on indigo_polecat/Pineapple


src/constants/accessories.js
```diff
@@ -775,3 +776,13 @@ export const MAGICAL_POWER = {
   special: 3,
   very_special: 5,
 };
+
+const recombableAccessories = uniqueAccessories;
+
+for (const accessory of Object.keys(recombableAccessories)) {
+  if (ACCESSORIES[accessory]?.allowRecomb == false) {
```
dukio: I would make this a `===`

dukio (changes requested):
Only a small thing then it's good to go for me (didn't test)

metalcupcake5:
image
dont merge until i do this

dukio:
Also would it be more grammatically correct to use "allowsRecomb" (with s?)

dukio (approved):
I'll approve this, but the accessories stats need to be revisited in the future...

image

Unique accessories is the number of unique accessories owned out of all the accessories that exist, correct.

Completion should be equivalent to "this person has this progress towards having every accessory upgraded to the max (recomb, enrichment, other stuff like max gemstones on artifact of power or pandora box challenges)"

Recombobulated should be, as it was before, "this person has recombobulated X accessories out of his total accessories owned that can be recombobulated", this PR changes this to "this person has X recombobulated accessoires out of all the accessories in existance that can be recombobulated" (so not just the owned ones)

MartinNemi03 (approved):
lgtm, didn't test

PR #1660, opened 2022-09-13
tell the browser whether to use light or dark mode

nstringham:
currently, we set everything to dark mode manually by making the background dark and the test light but by using the `color-scheme` CSS property we can tell the browser we are using dark mode and it will make the default be dark mode.


dukio (approved):
lgtm didnt test

MartinNemi03 (approved):

MartinNemi03:
lgtm, didn't test

PR #1666, opened 2022-09-14
feat: add networth

### LINKED ISSUE #1662: Profile networth
There's a npm package now that we could use https://www.npmjs.com/package/skyhelper-networth

DuckySoLucky:
## Description 

Closes #1662 
> Networth is not shown if player has inventory API off

## Examples
image
image
image


MattTheCuber:
Decided not to cache prices?

MattTheCuber:
Also purse and bank are kind of redundant

DuckySoLucky:
Didn't know u were able to do that with npm package, will update when I get home. Thanks

MattTheCuber:
> Didn't know u were able to do that with npm package, will update when I get home. Thanks

The package documentation shows how to do that

DuckySoLucky:
If I'm gonna be honest I only read part around getNetworth function ;-;

MattTheCuber (approved):
Glad to see that 2 platforms will now have a consistent networth calculation!

DuckySoLucky:
image

dukio (changes requested):
`stats.ejs` has extra code from equipment pr? how's this possible 🤔 

if you can fix it and upstream would be great!

src/lib.js
```diff
@@ -26,6 +27,11 @@ const hypixel = axios.create({
 });
 const parseNbt = util.promisify(nbt.parse);
 
+let prices = await getPrices();
```
dukio: I'm not 100% sure this is the best place for this... 

I have a feeling it should be similar to `src/scripts/update-bazaar.js`, opinions?

dukio (changes requested):
See https://github.com/DuckySoLucky/SkyCrypt/pull/10 + my comment

src/lib.js
DuckySoLucky: First I thought about doing it like that but that would just slow down stuff since it has to take values from database. You choose, doesn't really matter

dukio: I'll wait for the opinion of other @SkyCryptWebsite/developers

MartinNemi03 (approved):
lgtm, tested

PR #1668, opened 2022-09-14
fix: incorrect HOTM token

DuckySoLucky:
## Description 

Fixes a bug where it would say x/x instead of x/y
> x = player's unlocked tokens, y = maximum amount of tokens u can have

### Old
image

### New
image

### DIFF
--- a/src/constants/hotm.js
+++ b/src/constants/hotm.js
@@ -1421,6 +1421,7 @@ export const HOTM = {
   nodes: nodeClasses,
   items: [HotmStats, CrystalHollowsCrystals, HotmReset],
   powder_for_max_nodes: powderForMaxNodes,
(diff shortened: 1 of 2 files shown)


dukio:
A little explanation here for other reviewers:

The "golden" text was meant to indicate if the player was using ALL the tokens in his possession... so if the user has max 7 tokens unlocked and it's using them all, then golden text.
It was meant as a way to display if the user was "wasting" tokens by not using them; so no golden text if unused tokens.

But the truly correct way of using the golden (meaning a max setup) is to only give gold if the user **owns** and is using all tokens.

DuckySoLucky:
Fixed with better way in #1694, So I'm gonna close this

PR #1690, opened 2022-09-28
hotfix for profiles

DuckySoLucky:


src/lib.js
```diff
@@ -3267,93 +3230,56 @@ export async function getProfile(
 
       allSkyBlockProfiles.push(doc);
 
-      lastCachedSave = Math.max(lastCachedSave, doc.members[paramPlayer].last_save || 0);
+      lastCachedSave = Math.max(lastCachedSave, Date.now() || 0);
```
dukio: This will always be `Date.now()`, right?

src/constants/hotm.js
```diff
@@ -759,7 +759,7 @@ class PeakOfTheMountain extends Node {
     const output = [];
 
     for (let tier = 1; tier <= level; tier++) {
-      for (const [reward, qty] of Object.entries(rewards.potm[tier])) {
+      for (const [reward, qty] of Object.entries(rewards.hotm[tier])) {
```
dukio: this `potm` was correct, if it creates other errors those needs to be fixed

dukio (changes requested):
didn't test, reviewed code on github and left a couple comments

src/lib.js
DuckySoLucky: Yeah, it's useless will remove tomorrow

src/constants/hotm.js
DuckySoLucky: Yeah I'm aware as I said on discord they have updated the hotm tree. This fixes error so it's kinda a fix for it as well. Will update it tomorrow

dukio: image

Oh shit missed this in patch notes... welp found why then

src/lib.js
DuckySoLucky: Solved, didn't remove found better way

src/constants/hotm.js
DuckySoLucky: Solved

PR #1698, opened 2022-10-02
Fix for "User not found" when directly entering cute_name in URL

MartinNemi03:
This mainly happens when you try to directly get a currently not selected profile by it's `cute_name`. 

Examples: https://sky.shiiyu.moe/stats/MartinNemi03/Cucumber, https://sky.shiiyu.moe/stats/Oblivion239/Mango

_Thanks to Oblivion239 for reporting this bug!_

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -3362,7 +3362,11 @@ export async function getProfile(
       return;
     }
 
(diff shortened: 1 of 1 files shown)


dukio (approved):
lgtm didnt test

PR #1699, opened 2022-10-03
feat: add Accessories bag data and axes to farming tools

### LINKED ISSUE #657: Add axes to farming hoes section
some items such as the Pumpkin Dicer are not hoes but should be included in the farming hoes section

this section should probably also be renamed to farming tools

### LINKED ISSUE #1381: API accessory_bag_storage
See https://github.com/HypixelDatabase/HypixelTracking/commit/95debe446433d06d016c650304ecb1e38e382631#diff-618eebadb464b46962876276524fa5b544f7cf5f45b6b2277fd75266d88c1823R6094-R6121

tuning, unlocked slots, unlocked powers, selected power, bag upgrades purchased

DuckySoLucky:
## Description

Closes #657 
Closes #1381 

## Examples
image
image
image

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -1123,7 +1123,15 @@ export const getItems = async (
   output.accessory_rarities = accessoryRarities;
 
   output.weapons = allItems.filter((a) => a.categories?.includes("weapon"));
-  output.farming_tools = allItems.filter((a) => a.categories?.includes("farming_tool") || a.tag?.ExtraAttributes?.id == "MELON_DICER" || a.tag?.ExtraAttributes?.id == "PUMPKIN_DICER" || a.tag?.ExtraAttributes?.id == "COCO_CHOPPER" || a?.tag?.ExtraAttributes?.id == "FUNGI_CUTTER" || a?.tag?.ExtraAttributes?.id == "CACTI_KNIFE");
+  output.farming_tools = allItems.filter(
+    (a) =>
+      a.categories?.includes("farming_tool") ||
+      a.tag?.ExtraAttributes?.id == "MELON_DICER" ||
+      a.tag?.ExtraAttributes?.id == "PUMPKIN_DICER" ||
+      a.tag?.ExtraAttributes?.id == "COCO_CHOPPER" ||
+      a?.tag?.ExtraAttributes?.id == "FUNGI_CUTTER" ||
+      a?.tag?.ExtraAttributes?.id == "CACTI_KNIFE"
+  );
   output.mining_tools = allItems.filter((a) => a.categories?.includes("mining_tool"));
   output.fishing_tools = allItems.filter((a) => a.categories?.includes("fishing_tool"));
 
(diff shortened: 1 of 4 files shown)


src/lib.js
```diff
@@ -1123,7 +1123,15 @@ export const getItems = async (
   output.accessory_rarities = accessoryRarities;
 
   output.weapons = allItems.filter((a) => a.categories?.includes("weapon"));
-  output.farming_tools = allItems.filter((a) => a.categories?.includes("farming_tool"));
+  output.farming_tools = allItems.filter(
+    (a) =>
+      a.categories?.includes("farming_tool") ||
```
dukio: my initial idea when doing the categories was to find a place in lib.js where we could add a bunch of IFs like
`if item.id === "MELON_DICER", item.categories.push("farming_tool")`
so that if we need to print farming tools in different placed we dont have to rememeber all the exceptions.

would you mind looking for a solution like this instead? could be useful in the future for other things too (not just farming tools maybe, but other categories too)

views/stats.ejs
```diff
@@ -1016,7 +1016,18 @@ const metaDescription = getMetaDescription()
...
+
+            <span data-tippy-content="
+            <span style='color: var(--§8);' class='grey-text'>Unlocked Reforges</span><br><br>
+              <% for (const reforge of calculated.accessory_bag_storage.unlocked_powers || []) { %>
+                <span style='color: var(--§8);' class='grey-text'>● <span> <span style='color: var(--§f);' class='grey-text'> <%= (reforge.charAt(0).toUpperCase() + reforge.slice(1)) %></span><br>
+              <% } %>
+              ">
+              <span class="stat-name">Selected Power: </span><span class="stat-value"><%= calculated.accessory_bag_storage.selected_power ? (calculated.accessory_bag_storage.selected_power.charAt(0).toUpperCase() + calculated.accessory_bag_storage.selected_power.slice(1)) : 'None'%></span>
+            </span><br>
+            <% max = calculated.accessory_bag_storage.bag_upgrades_purchased >= 99 ? 'golden-text' : '' %>
```
dukio: the 99 I would make it a CONSTANT, better to not hardcode values outside constants folder

dukio (commented):
Couple things to review. In the future (or if you have time/will now) would be nice to also include the missing reforges (if they are not too many, maybe as a normal text list (`foo, bar and muu`)) + a count "unlocked 10 / 99"

src/lib.js
DuckySoLucky: Updated in latest commit

views/stats.ejs
DuckySoLucky: Done in latest commit

dukio (changes requested):
Does this still includes treecaps for you too?

dukio:
image
http://localhost:32464/stats/Refraction/Apple#Skills

For some weird reason it also puts treecapitators in... it does it with any of these conditions `itemId === "MELON_DICER" || itemId === "PUMPKIN_DICER" || itemId === "COCOA_CHOPPER"` but not if you do like `itemId === "xxx"` which is even weirder... 

There must be something that I'm missing o-ò

DuckySoLucky:
> Does this still includes treecaps for you too?

Fixed, still not sure what was the cause :/

DuckySoLucky:
There's new PR for accessories

PR #1651, opened 2022-09-11
feat: add bestiary

DuckySoLucky:
## Description 

Closes #1461 

## Examples
image
image
image

PR #1701, opened 2022-10-03
Fix DOM issues

dukio:
This PR fixes some DOM issues I found trying to format our final output html with prettier:
- `<source />` tags are self closing, not `</source>`
- some classes would end with double quote `class="..."">`
- an extra `</span>`
- a `</p>` that was being closed way too late

Note for myself, look at #1651 after this is merged (1651 should upstream)


MartinNemi03 (approved):
lgtm, didn't test

PR #1718, opened 2022-10-11
Switch to PNPM

dukio:
It's faster 🚀 

+bonus
- Now when running `pnpm dev` it will automatically check that you have the correct dependencies installed (no more need to `npm ci` when checking out a PR or after pulling commits!)

This is now possible because it's simply so fast you won't even notice it ...and it doesn't randomly break like with `npm i`


dukio:
@WarpWing can you tell me what should be done to the Dockerfile so it works with pnpm? (I have no clue how to do/test it)

MartinNemi03 (approved):
Definitely seems like a interesting tool to use.. Once merged, I'll also update the autoupdate scripts.

WarpWing:
Every time the majority of devs approve a PR but doesn't merge it, a kitten gets killed and it's their collective fault.

PR #1731, opened 2022-10-22
fix: Social Experience not being shared between coop members

### LINKED ISSUE #1519: Social Skill Leveling
The Social level on the SkyCrypts website doesn't match with my ingame Social level.
I've done some research on how the social skill leveling works because it felt kind of odd for me and found out that Hypixel stores only the player's social xp, that means if you are in a coop and someone visits you through your coopmate (/visit coopmate) then the XP doesn't add under your profile in SkyBlock Public API, but rather adds under your coopmate's profile in the API.
So you have to loop through every coop member until you get all the Social XP from the specific coop profile
unknown (8)
unknown (7)

DuckySoLucky:
## Description

Closes #1519

image

> After

image

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -1386,7 +1386,7 @@ export const getItems = async (
   return output;
 };
 
-async function getLevels(userProfile, hypixelProfile, levelCaps) {
+async function getLevels(userProfile, hypixelProfile, levelCaps, profileMembers) {
   const output = {};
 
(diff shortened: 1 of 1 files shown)


MartinNemi03 (approved):
i hate we have to do it this way, but who would have expected hypixel would ever do it this way.. lgtm i guess..

PR #1694, opened 2022-09-30
Optimizations and updates

DuckySoLucky:
## Description

Cleaned some parts of code
Updated pets, added ability to use profile stats for pet stats (Golden Dragon gold Collection)
Fixed `userProfile.tutorial is not iterable` Error [Click for Example](https://sky.shiiyu.moe/stats/beb5b536b6f1491e9c3b218b5a19063c/5bd6d546cab2431b93d672a13ef2e744) 

Closes #1689 
Closes #1572 
Closes #986 
Closes #1199

PR #1747, opened 2022-10-30
Custom resources caching

MartinNemi03:
Custom Resources are famous for their ridiculous load times that slow down startup times, so we went ahead and implemented caching for custom resources! 

Upon the first load, a cache with pack config hashes (for checking changes) and loaded resource packs will be created in your local /cache folder. From there, every startup it will check if pack configs have changed and if not, just load the packs from cache!

Included with this PR is a optimization change (by @DuckySoLucky) from #1694! 
_...And some new debug stats for how many resources needed to be checked with how long it took I guess._

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -28,6 +28,7 @@ import { mongo, db } from "./mongo.js";
 import sanitize from "mongo-sanitize";
 import * as helper from "./helper.js";
 import * as constants from "./constants.js";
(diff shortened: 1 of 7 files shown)


MartinNemi03:
The only thing missing right now is displaying debug stats only in dev mode, but it's pretty much prepared for that.

PR #1752, opened 2022-10-31
fix: abiphone being undefined and Kuudra core upgrades

DuckySoLucky:
## Description

If player had an abicase and no contacts inside of abiphone it would cause an error

Example: https://sky.shiiyu.moe/stats/b44d2d5272dc49c28185b2d6a158d80a/3a0a7a5473cd4491a8b9c523ee530ec2


MartinNemi03 (approved):
lgtm, didn't test

PR #1756, opened 2022-11-03
fix: Pet Score MF Symbol

DuckySoLucky:
## Description

Fixes https://canary.discord.com/channels/738971489411399761/738990246825427084/1036219139108380752
> TL;DR Calculation would put pointing arrow to closest value even without meeting requirements

Example: https://sky.shiiyu.moe/stats/dukioooo/Raspberry#Pets

image

> After

image

### DIFF
--- a/views/stats.ejs
+++ b/views/stats.ejs
@@ -1191,8 +1191,10 @@ const metaDescription = getMetaDescription()
               <table>
               <% 
               const entries = Object.entries(petRewards)
-              let reduced = Object.keys(petRewards).reduce((prev, curr) => Math.abs(curr - petScore) < Math.abs(prev - petScore) ? curr : prev)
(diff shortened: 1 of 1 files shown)


MartinNemi03 (approved):
lgtm, didn't test

Shiiyu (changes requested):
Instead of such a crude workaround, I'd suggest it to be changed to findLast() instead, as discussed earlier, requesting changes here so it isn't accidentally merged from Martin's approval.

DuckySoLucky:
> Instead of such a crude workaround, I'd suggest it to be changed to findLast() instead, as discussed earlier, requesting changes here so it isn't accidentally merged from Martin's approval.

Done but different way, `.findLast()` doesn't seem to work. Check discord for more information

Shiiyu (approved):
LGTM

PR #1771, opened 2022-11-10
fix: Collection amount

DuckySoLucky:
## Description

Fixes Collections above integer limit not being shown

image

> After

image


MartinNemi03:
The Farming XP bar should be working as intended, since it should be filled up when you reach your level cap since you can't progress further until you unlock the next level cap in Jacob's house. That's mainly why level cap of Farming also depends on what that perk is leveled at.

If there's something wrong with current approach and/or if there really is an issue, please describe it further.

MartinNemi03:
Also after checking the collection thing, it's incorrectly noted as an "Collections above integer limit not being shown", since that should also be intended behavior since it only shows the collection count only collected by the user in a coop, not the total of the coop. You can actually check that by hovering on the collection. 
In this case, I would rather suggest changing "Amount:" to "Total:" and by extension this wouldn't be a bugfix, but just a change of what we display.

PR #1777, opened 2022-11-16
feat: Hypixel+ 0.17.0

ic22487:
Adds the Hypixel+ Resource Pack to SkyCrypt


MartinNemi03 (changes requested):
Sorry for taking so long before looking into this PR.
From first look, this looks very good, good job on the RP!
Only thing is that I would like to suggest changing the url and version in config.json into the same format as the other ones.
_Also I'm not sure if matching priority with other pack could potentially break things, but that's not that big._

ic22487:
> Only thing is that I would like to suggest changing the url and version in config.json into the same format as the other ones.

My pack's version number typically coincides with the skyblock version, then a pack-specific patch version (hence the 2 different numbers) - would the sb version be ok to put in the name instead of the version number, or would that mess things up in the long-term?

MartinNemi03 (approved):
lgtm, tested

PR #1792, opened 2022-12-04
feat: add Belts

DuckySoLucky:
## Description

Adds dojo belts to the accessories

## Examples
image
image


ic22487:
Arent the dojo belts equipment, not accessories?

PR #1795, opened 2022-12-09
fix: Error caused by enrichments

DuckySoLucky:
## Description

Fixes `Cannot read properties of null (reading 'toLowerCase')` caused by `getEnrichments()` function
Example: https://sky.shiiyu.moe/stats/Milo0209
Also refactor to social experience leveling not really required but it wasn't really safe, it could have caused error. So that's fixed now


MartinNemi03 (approved):
lgtm, didn't test

PR #1799, opened 2022-12-11
fix: wrong max Experience enchantment level

DuckySoLucky:
## Description

Changed Experience Enchantment level from IV to V, Experience V book was added in Jerry Island Update in the Ice Essence shop

## Example
image


MartinNemi03 (approved):
lgtm, didn't test

PR #1800, opened 2022-12-11
fix: incorrect Ice and Snow Minion max tier

DuckySoLucky:
## Description

Fixed incorrect Ice and Snow Minion max tier, added in Winter Island Update

image
> After

image


MartinNemi03 (approved):
lgtm, didn't test

PR #1814, opened 2023-01-01
feat: Commissions milestone max

DuckySoLucky:
## Description

Adds https://canary.discord.com/channels/738971489411399761/738990231348314123/1059094726562824222
>make commissions turn gold when at 750

image

> After
image

### DIFF
--- a/src/constants/hotm.js
+++ b/src/constants/hotm.js
@@ -1445,5 +1445,3 @@ export const PRECURSOR_PARTS = {
 };
 
 export const COMMISSIONS_MILESTONE = 6;
-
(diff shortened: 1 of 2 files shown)


MartinNemi03:
Do we "gold" other things that come from achievements as well? 
Since you could possibly get max commissions on a profile that hasn't done the max milestone.

DuckySoLucky:
> Do we "gold" other things that come from achievements as well? Since you could possibly get max commissions on a profile that hasn't done the max milestone.

No, and only thing we get from achievements are secrets so yeah.. Not sure tbh

MartinNemi03 (approved):
lgtm, because there's.. nothing.

PR #1835, opened 2023-01-22
fix: slothpixel API not having all accessories

DuckySoLucky:
## Description

Fixed talismans not being shown in missing talismans area due to SlothPixel API being outdated
Also fixes `ARCHAEOLOGIST_COMPASS` accessory


DuckySoLucky:
Closing since slothpixel API fixed this

PR #1836, opened 2023-01-24
refactor: missing accessories system

DuckySoLucky:
## Description

Uses `/items` endpoint to get accessories instead of hard coding every one of them
cc @MattTheCuber

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -1138,8 +1138,8 @@ export const getItems = async (
     }
 
     if (id.startsWith("PARTY_HAT_CRAB")) {
-      const CRAB_HAT = accessories.find((a) => helper.getId(a) == `${id}`);
-      const CRAB_HAT_ANIMATED = accessories.find((a) => helper.getId(a) == `${id}_ANIMATED`);
+      const CRAB_HAT = accessories.find((a) => helper.getId(a) == `PARTY_HAT_CRAB`);
+      const CRAB_HAT_ANIMATED = accessories.find((a) => helper.getId(a) == `PARTY_HAT_CRAB_ANIMATED`);
 
       if (CRAB_HAT && CRAB_HAT_ANIMATED) {
         CRAB_HAT.isUnique = false;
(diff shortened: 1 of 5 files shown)


src/lib.js
```diff
@@ -1147,8 +1147,8 @@ export const getItems = async (
       }
     }
 
-    if (id in constants.ACCESSORY_UPGRADES) {
-      const accessoryUpgrades = constants.ACCESSORY_UPGRADES[id];
+    if (constants.getUpgradeList(id) !== undefined) {
```
MattTheCuber: `!== undefined` is redundant (used several places). Unless you want to exclude other falsy values, but I don't think you do

DuckySoLucky: > `!== undefined` is redundant (used several places). Unless you want to exclude other falsy values, but I don't think you do

Just realized it returns boolean, I'm dumb ;-;

DuckySoLucky:
Also fixes inactive accessories not being shown
> Before

image

> After

image

DuckySoLucky:
Found bug, cba to fix it today

DuckySoLucky:
> Found bug, cba to fix it today

Fixed

DuckySoLucky:
> Also fixes inactive accessories not being shown

Made it better, now it shows accessories which aren't in accessory bag or aren't unique as well
> Before

image
> After

image

DuckySoLucky:
Added `Location` field to inactive accessories so players can see where the accessory is located at
image

MattTheCuber:
> Made it better, now it shows accessories which aren't in accessory bag or aren't unique as well

What about accessories you don'teet requirements for like Reaper Orb?

DuckySoLucky:
> > Made it better, now it shows accessories which aren't in accessory bag or aren't unique as well
> 
> What about accessories you don'teet requirements for like Reaper Orb?

~~Good idea, gonna add that. Most likely tomorrow~~
Edit: Nothing about that, SkyCrypt doesn't like that for some reason..

MartinNemi03 (approved):
lgtm, tested on only a few profiles

PR #1838, opened 2023-01-27
fix: Farming Level capped at 50 if skills API off

DuckySoLucky:
## Description

Fixes https://canary.discord.com/channels/738971489411399761/748625263893545042/1066876666619646064
> `Math.max()` can be used safely here since hypixel returns true level from achievements

image
> After

image

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -94,7 +94,9 @@ function getXpByLevel(uncappedLevel, extra = {}) {
 
   /** the level that this player is caped at */
   const levelCap =
-    extra.cap ?? Math.max(uncappedLevel, constants.DEFAULT_SKILL_CAPS[extra.skill]) ?? Math.max(...Object.keys(xpTable).map((a) => Number(a)));
+    extra.cap ??
+    Math.max(uncappedLevel, constants.DEFAULT_SKILL_CAPS[extra.skill]) ??
+    Math.max(...Object.keys(xpTable).map((a) => Number(a)));
 
   /** the maximum level that any player can achieve (used for gold progress bars) */
   const maxLevel = constants.MAXED_SKILL_CAPS[extra.skill] ?? levelCap;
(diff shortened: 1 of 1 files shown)


WarpWing (changes requested):
Could we use OR statements here instead? Not to sound pedantic but instead, you could use || instead of ??. So it would look like

```js
const levelCap = extra.cap ||  (uncappedLevel > (extra.skill in constants.DEFAULT_SKILL_CAPS ? constants.DEFAULT_SKILL_CAPS[extra.skill] : uncappedLevel)) || Math.max(...Object.keys(xpTable));
```

MartinNemi03:
> Could we use OR statements here instead? Not to sound pedantic but instead, you could use || instead of ??. So it would look like
> 
> ```js
> const levelCap = extra.cap ||  (uncappedLevel > (extra.skill in constants.DEFAULT_SKILL_CAPS ? constants.DEFAULT_SKILL_CAPS[extra.skill] : uncappedLevel)) || Math.max(...Object.keys(xpTable));
> ```

Not really sure what would the benefit be there, maybe once it would be prettified it would be better, but from this it doesn't look any better, at least only from the looks of it.

DuckySoLucky:
> > Could we use OR statements here instead? Not to sound pedantic but instead, you could use || instead of ??. So it would look like
> > ```js
> > const levelCap = extra.cap ||  (uncappedLevel > (extra.skill in constants.DEFAULT_SKILL_CAPS ? constants.DEFAULT_SKILL_CAPS[extra.skill] : uncappedLevel)) || Math.max(...Object.keys(xpTable));
> > ```
> 
> Not really sure what would the benefit be there, maybe once it would be prettified it would be better, but from this it doesn't look any better, at least only from the looks of it.

https://canary.discord.com/channels/738971489411399761/820755886262976553/1069171051268952166
We talked about it, `??` would be better in this case

MartinNemi03 (approved):
lgtm, didn't test

PR #1841, opened 2023-01-28
refactor: getMinions and getMinionSlots optimization 

WarpWing:
Optimized getMinionSlotsremoving by unnecessary sorting of the MINION_SLOTS keys and used the Array.prototype.reduce method to count the number of unique minions efficiently. Also removed weird comparisons and index variable, replaced with indexOf().


DuckySoLucky:
Front page where

WarpWing:
> Front page where

We are not worthy and that makes me sadge. (I'll sneak it in with the next big optimization)

MartinNemi03:
> > Front page where
> 
> We are not worthy and that makes me sadge. (I'll sneak it in with the next big optimization)

You know that I actually look into what changes are made?

WarpWing:
> > > Front page where
> > 
> > 
> > We are not worthy and that makes me sadge. (I'll sneak it in with the next big optimization)
> 
> You know that I actually look into what changes are made?

Stop. Don't do that. Competency is overrated.

WarpWing:
DO NOT MERGE THIS CODE BREAKS

PR #1843, opened 2023-01-29
We do a little bit of async

WarpWing:


DuckySoLucky:
Hi daddy

MartinNemi03 (commented):
lgtm, although i don't really see why some things changed in regards to async
i thought we want to run all master functions before the site itself runs (so that the scripts are all done before we fork)

PR #1845, opened 2023-01-31
refactor: Included Hypixel API Key and optimized Docker Compose file

WarpWing:
Hi all,

I decided to do some refactoring with the Docker Compose file to allow an automatic restart and updated the version building upon Hannah's work.  However, I want to quickly touch on my reasoning for adding blank text. During my (admittedly little) time testing SkyCrypt, I found it quite burdensome to find a way to include a CI/CD pipeline that properly injects a Hypixel API key. So I'm thinking for testing, we could put placeholder text so at least we can run E2E testing. I think I found a way to inject an actual key using an ad hoc script that injects the key into credentials.json after generation. But for now for accelerated E2E testing, it's a do-able solutions.


WarpWing:
Closing this and reworking it for later

PR #1853, opened 2023-02-07
Account For Finnegan Perk and Correct Medal Brackets

### LINKED ISSUE #1852: Incorrect Jacob's Medal Brackets
**Describe the bug**
The website calculates gold/silver/bronze medals incorrectly for Jacob's contests. For example with a gold medal, the actual formula is `position <= participants * 0.05 + 1`. That `+1` is the discrepancy here, it's unclear whether Hypixel simply rounds up, or adds one, but it's definitely not left alone. You can see an example of why this was figured out to be the case here:

image

Even though this player didn't reach the amount required for gold, they still got gold because of that plus one. The theory is that this ensures 1 person will get gold if a contest only has, say, 10 participants.

Also, there's a second issue here. The new Finnegan mayor introduced a `claimed_medal` property to the API response because of the perk to expand medal brackets. This value should supersede all calculations done manually.

image

**Profile/Location**
All profiles could be affected, but note this is an edge case, so numbers are only slightly wrong for some people. With that in mind, an example:

https://sky.shiiyu.moe/stats/Bankhier/Papaya#Skills
image
(What the numbers should be to the best of my knowledge and that of the Elite Farmers)
image

**Additional Info**
N/A

**Expected behavior**
Just that these numbers are counted correctly as described above.

**Screenshots**
(Put them above already, hopefully that's fine)

- OS: Windows 10
- Browser: Firefox
- Version 109.0.1

**Additional context**
The implementation that I use can be found here, but I will open a pull request with these changes in a moment anyway. https://github.com/ptlthg/EliteWebsite/blob/main/src/lib/data.ts#L530

ptlthg:
Fixes #1852

Adjusts calculations to the ones as described in the issue, and ensures the `claimed_medal` property is respected.

Note that `placing.percentage` is unused now, but I'm unsure if that is planned on being used elsewhere, so I left it.

### DIFF
--- a/src/lib.js
+++ b/src/lib.js
@@ -1906,6 +1906,10 @@ export async function getStats(
       const date = `${contestName[1]}_${contestName[0]}`;
       const crop = contestName.slice(2).join(":");
 
+      if (data.collected < 100) {
+        continue; // Contests aren't counted in game with less than 100 collection
+      }
+
       farming.crops[crop].contests++;
       farming.crops[crop].attended = true;
       if (farming.crops[crop].personal_best < data.collected) {
@@ -1925,19 +1929,25 @@ export async function getStats(
       if (contest.claimed) {
         placing.position = data.claimed_position || 0;
         placing.percentage = (data.claimed_position / data.claimed_participants) * 100;
+        const participants = data.claimed_participants;
 
-        if (placing.percentage <= 5) {
+        // Use the claimed medal if it exists and is valid
+        // This accounts for the farming mayor increased brackets perk
+        // Note: The medal brackets are the percentage + 1 extra person
+        if (contest.claimed_medal === "bronze" || contest.claimed_medal === "silver" || contest.claimed_medal === "gold") {
+          contest.medal = contest.claimed_medal;
+        } else if (placing.position <= participants * 0.05 + 1) {
           contest.medal = "gold";
-          farming.total_badges.gold++;
-          farming.crops[crop].badges.gold++;
-        } else if (placing.percentage <= 25) {
+        } else if (placing.position <= participants * 0.25 + 1) {
           contest.medal = "silver";
-          farming.total_badges.silver++;
-          farming.crops[crop].badges.silver++;
-        } else if (placing.percentage <= 60) {
+        } else if (placing.position <= participants * 0.6 + 1) {
           contest.medal = "bronze";
-          farming.total_badges.bronze++;
-          farming.crops[crop].badges.bronze++;
+        }
+
+        // Count the medal if it exists
+        if (contest.medal) {
+          farming.total_badges[contest.medal]++;
+          farming.crops[crop].badges[contest.medal]++;
         }
       }
 
--- a/src/lib.js
+++ b/src/lib.js
@@ -1934,7 +1934,11 @@ export async function getStats(
         // Use the claimed medal if it exists and is valid
         // This accounts for the farming mayor increased brackets perk
         // Note: The medal brackets are the percentage + 1 extra person
-        if (contest.claimed_medal === "bronze" || contest.claimed_medal === "silver" || contest.claimed_medal === "gold") {
+        if (
+          contest.claimed_medal === "bronze" ||
+          contest.claimed_medal === "silver" ||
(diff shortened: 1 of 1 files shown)


DuckySoLucky (approved):
lgtm, tested

ptlthg:
I admittedly didn't test these changes, I don't have a MongoDB instance and didn't want to set that up for just a small tweak like this. I *think* it should be fine, but yeah, be sure to check it out before merging. Thanks!

DuckySoLucky:
Tested on 40 profiles

## Example
#### DeathStreeks

>Before

image

> After

image

ptlthg:
Looks good to me, Deathstreeks has a lot of contests with only a handful of crops collected, which are no longer counted with this change. Probably one of the more extreme examples there.

<details>
  <summary>Just some of the contests</summary>
  
  image
</details>

MartinNemi03 (approved):
lgtm, tested

DuckySoLucky:
@ptlthg Can you please merge https://github.com/ptlthg/SkyCrypt/pull/1?

ptlthg:
Merged it, I did run prettier before but I thought the style issues were from elsewhere, I probably just missed a step.

ptlthg:
I noticed earlier today that the contest participation amounts weren't quite lining up with what's displayed in game with these changes. I believe I've identified the problem, and it looks like Hypixel is actually counting it wrong.

Take my contests for example:
image
If you compare that to the amount the site will give with these proposed changes, it only displays `521` contests. We're seemingly missing 3 contests that should be counted.

After some investigating, I've discovered that a variety of "ghost" contests are to blame. When you participate in two contests at the same time, they're both counted as you'd expect, but you can only claim rewards for one of them. The in-game menu also confusingly displays two contests, but only one is claimable, and the other is a duplicate which is taking the place of the other crop you participated in.

<details>
  <summary>See an example of the duplicated entry</summary>
  
  image
  image
</details>

However, these "ghost" contests are already treated like any other contest, and the issue isn't with all of them in general. The problem comes from when this "ghost" contest has less than 100 collection, Hypixel counts these as a participation while the website doesn't.

With that in mind, we can easily correct for this if all we want is to match what Hypixel says in game, but that's up to you all if you want to do that. While counting these ghost contests even if they have less than 100 collection is what Hypixel is doing, it's most likely a bug. (Also note, they're only counted if another crop is above 100 collection during that contest)

If you all think they should be counted to match what Hypixel shows, I'd be happy to add that patch to this PR.

MartinNemi03:
> I noticed earlier today that the contest participation amounts weren't quite lining up with what's displayed in game with these changes. I believe I've identified the problem, and it looks like Hypixel is actually counting it wrong.
> 
> Take my contests for example: image If you compare that to the amount the site will give with these proposed changes, it only displays `521` contests. We're seemingly missing 3 contests that should be counted.
> 
> After some investigating, I've discovered that a variety of "ghost" contests are to blame. When you participate in two contests at the same time, they're both counted as you'd expect, but you can only claim rewards for one of them. The in-game menu also confusingly displays two contests, but only one is claimable, and the other is a duplicate which is taking the place of the other crop you participated in.
> See an example of the duplicated entry
> 
> image image
> 
> However, these "ghost" contests are already treated like any other contest, and the issue isn't with all of them in general. The problem comes from when this "ghost" contest has less than 100 collection, Hypixel counts these as a participation while the website doesn't.
> 
> With that in mind, we can easily correct for this if all we want is to match what Hypixel says in game, but that's up to you all if you want to do that. While counting these ghost contests even if they have less than 100 collection is what Hypixel is doing, it's most likely a bug. (Also note, they're only counted if another crop is above 100 collection during that contest)
> 
> If you all think they should be counted to match what Hypixel shows, I'd be happy to add that patch to this PR.

I would consider not counting them, if it's a very likely bug.

ptlthg:
> I would consider not counting them, if it's a very likely bug.

I agree with not counting them, but it's just a matter of whether the "Contests Attended" value should exactly match the in game GUI or not. I did submit a Hypixel bug report, but we'll see if that does anything.

If they should be left uncounted, this PR is ready for merge still.

PR #1870, opened 2023-02-15
fix: SkyBlock Level bar width

DuckySoLucky:
## Description

Fixes https://canary.discord.com/channels/738971489411399761/738990246825427084/1075509107882209312
> Width of SkyBlock Level bar isn't correct on mobile devices

## Examples

> Before

image

> After

image
image

### DIFF
--- a/public/resources/scss/stats.scss
+++ b/public/resources/scss/stats.scss
@@ -94,8 +94,11 @@
 
     &[type="skyblock_level"] {
       margin-bottom: 25px;
-      padding-right: 24px;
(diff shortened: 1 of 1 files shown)


nstringham:
In the screenshot there is not enough space between the level and the taming

MartinNemi03:
> In the screenshot there is not enough space between the level and the taming
> 
> 

Changed by removing the margin that was higher than the one making the space between them

image

MartinNemi03 (approved):
lgtm, tested

PR #1872, opened 2023-02-16
feat: add leaderboard positions

DuckySoLucky:
## Description

Fixes https://canary.discord.com/channels/738971489411399761/738990246825427084/1075763157257162793
> Add leaderboard positions to missing progress bars (dungeons & SkyBlock level)

## Examples
image
image


MartinNemi03 (approved):
lgtm, didn't test

src/helper.js
MartinNemi03, line 1: i believe we should split the redis things from helper.js later, because now when accessing helper, it will start the redis client

MartinNemi03 (changes requested):
The level leaderboards don't even exist at the moment (checked the site's redis instance) and even then, I would not make seperate leaderboards for XP and Level at the same time, since level can be easily calculated afterwards and for ranking purposes XP is much better and more precise

MartinNemi03 (approved):
LGTM, didn't test, but the code looks fine at first glance

PR #1877, opened 2023-02-17
fix: error thrown if guild isn't in ur db (first search only)

DuckySoLucky:
## Description

Fixes https://canary.discord.com/channels/738971489411399761/738990246825427084/1076141595180670997
> TL;DR If u searched for guild which wasn't in your database it would throw an error first time
> Also cleaned code


metalcupcake5 (approved):
tested

PR #1885, opened 2023-02-24
fix: axes missing in farming tools

DuckySoLucky:
## Description

Fixes https://discordapp.com/channels/738971489411399761/739316781096763485/1036042056579166248
> Axes missing in farming tools

Wasn't sure how to make this without hard coding every item so I hard coded every enchantment which can be applied to farming tool, not the best way but better than hard coding every item

image

> After 

image


MartinNemi03 (approved):
lgtm, didn't test

PR #1890, opened 2023-03-01
fix: Magical Power being 2x

DuckySoLucky:
## Description

Magical Power is 2x of what it's suppose to be

## Examples

> Before

image

> After

image


MartinNemi03 (approved):
lgtm, didn't test

PR #1895, opened 2023-03-13
Farming weight

metalcupcake5:
adds farming weight to farming section
image

also updates my front page message


src/lib.js
```diff
@@ -3491,6 +3492,158 @@ async function getForge(userProfile) {
...
+
+  if (userProfile?.farming) {
+    let double_drops = userProfile.farming?.perks?.double_drops ?? 0;
+    output.bonuses.double_drops = {
+      double_drops: double_drops,
+      weight: double_drops * 2,
+    };
+    bonus += double_drops * 2;
+
+    let gold_medals = userProfile.farming?.total_badges?.gold ?? 0;
```
ptlthg: Note that the gold medal counter the site has is bugged, so the weight total might differ slightly. #1853 fixes this.

src/lib.js
```diff
@@ -3491,6 +3492,158 @@ async function getForge(userProfile) {
...
+        level: 0,
+        weight: 0,
+      },
+    },
+  };
+
+  const crops = {
+    CACTUS: {
+      name: "Cactus",
+      weight: 169_389.33,
```
ptlthg: All the weight values should probably be moved into a constants file for easily adjusting them in the future.

views/stats.ejs
```diff
@@ -1655,6 +1655,22 @@ const metaDescription = getMetaDescription()
             <% const farming = calculated.farming;
           if (farming.contests.attended_contests > 0) {%>
             <p class="stat-raw-values">
+              <span class="stat-name">Farming Weight: </span><span class="stat-value">
```
ptlthg: Should this only be shown if the player has participated in a contest, as it is now? Their farming weight will surely be low if they haven't, but still. The same with pelts, actually.

src/lib.js
```diff
@@ -3491,6 +3492,158 @@ async function getForge(userProfile) {
...
+    WHEAT: {
+      name: "Wheat",
+      weight: 100_000,
+    },
+  };
+
+  if (userProfile?.collections) {
+    let weight = 0;
+    for (const [name, crop] of Object.entries(crops)) {
+      let total = userProfile.collections[name].amount;
```
ptlthg:
```suggestion
      let total = userProfile.collections[name]?.amount ?? 0;
```
Added a fallback to zero, as I'm not sure if locked collections would still appear in here.

src/lib.js
MartinNemi03: I would also consider moving farming weight calculation to `weight/farming-weight.js`

ptlthg:
Before this can be merged, there's new weight values for the following after the Garden 1.5 update.

	pumpkin: 90_066.27,
	melon: 450_324.6,
	mushroom: 90_178.06,
	cocoa: 267_174.04,
	cactus: 177_254.45,
	
Axe crops were nerfed a tiny bit because of mossy fermento helm being better than lantern helm, and the Fungi Cutter was patched so mushroom weight is almost doubled now.

I do agree that this should all be moved into its own weight file like the other weight systems have, though.

**Edit:** Could also consider periodically fetching https://elitebot.dev/api/weights to get the current values, but they won't change often.

metalcupcake5:
would it be possible to create an NPM package so its just a simple package update?

views/stats.ejs
metalcupcake5: added

src/lib.js
metalcupcake5: added

ptlthg:
> would it be possible to create an NPM package so its just a simple package update?

I can definitely look into that for the future. I just have no free time right now, so it'll probably have to wait until summer.

This all looks good to me now though, thanks for adding everything!

Do I get a prestigious mention with the other weight systems in the about modal, though?
image

PR #1901, opened 2023-03-18
feat: add `Crystal Hollows` to Area Names

DuckySoLucky:
## Description

Adds `Crystal Hollows` to area names, somehow missed that on PR which I made like month ago :shrug:


MartinNemi03 (approved):
LGTM, thankfully it was covered by the fallback

PR #1916, opened 2023-03-30
fix: Skyblock Level maxLevel from 386 to 500

axlecoffee:
## Description

https://cdn.discordapp.com/attachments/738990246825427084/1091086444153544744/image.png

The maximum SkyBlock Level is SkyBlock Level 500.
https://wiki.hypixel.net/SkyBlock_Levels


DuckySoLucky (changes requested):
LoganLoganLogan has done calculation, max level at this moment is 398 so we should use that value instead of 500 in my opinion

https://canary.discord.com/channels/738971489411399761/1091086444321329232/1091521633228951582

metalcupcake5:
closing in favor of #2049

PR #1928, opened 2023-04-11
fix: valuable items being hidden

DuckySoLucky:
## Description

Fixes https://canary.discord.com/channels/738971489411399761/738990246825427084/1095312418915233844

image
> After

image


metalcupcake5 (approved):
do we still want to limit bonemerangs?

DuckySoLucky:
> do we still want to limit bonemerangs?

They're out of meta and like nobody uses them so there's no point

PR #1944, opened 2023-05-03
feat: member uuids are now filtered seperately

MartinNemi03:
mainly so we can filter out unconfirmed coop members


DuckySoLucky (approved):
lgtm, didn't test

MartinNemi03:
tested loading profiles on like 90 different profiles, works fine, so lgtm as well

PR #1957, opened 2023-06-20
Feat vampire slayer

DuckySoLucky:
## Description

Adds Vampire slayer which also fixes `Cannot convert undefined or null to object` error


MartinNemi03 (approved):
lgtm, didn't test

PR #2068, opened 2023-10-28
refactor: remove `Math.max()` for skyblockLevelCap

DuckySoLucky:
## Description

Does https://discord.com/channels/738971489411399761/738990246825427084/1167899407690240110
> New max level exists, top player alr reached it so this is useless (and there's 0% chance of us updating that value after every update drops so yeah..)


MartinNemi03 (approved):
LGTM, didn't test

PR #2140, opened 2023-12-30
perf: improve `getTexture()` up to 5x 

DuckySoLucky:
## Description

Improved performance by quite a lot by implementing caching, Sets, early exits out of for loop and early returns. Also changes some texture to use SkyBlock ID instead of texture value

## Preview

> `/stats/DeathStreeks/Blueberry` (100 tests)

Before: `7,864` - `14,442` ms
After: `1,208` - `1,985` ms

Improvement: `650.99337%` - `727.55667%`

### Known Issues

HOTM tree texutres are currently broken, this is intentional. I'm planning to rewrite HOTM tree fully, mostly because it takes around second and half to obtain textures for HOTM and the reason for that is because it does lore parsing which isn't really a good way to do it.


Zickles:
Lgtm

metalcupcake5 (approved):

PR #2158, opened 2024-01-09
Fixed Typos

AgedRob1n:
Fixed typos in CONTRIBUTING.md and README.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
-## Contributing to the SkyCrypt
+## Contributing to SkyCrypt
 
-Before contributing to SkyCrypt, make sure you install the development environment first. If you have trouble building SkyCrypt or have any development questions, please don't hesistate to contact us on [Discord](https://discord.gg/cNgADv2kEQ)!
+Before contributing to SkyCrypt, make sure you install the development environment first. If you have trouble building SkyCrypt or have any development questions, please don't hesitate to contact us on [Discord](https://discord.gg/cNgADv2kEQ)!
 
 ### Prerequisite Software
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -31,13 +31,13 @@ Before contributing to SkyCrypt, make sure you install the development environme
...
-6. (optional) If you are not using the default Redis port or you are using Redis remotely, you can configure the Redis URL with the `redisUrl` field in `credentials.json`. Also `discord_webhook` if you want to send error remotely, useful in production to detect bugs.
-7. Making sure your Mongo and Redis instances are running, run `pnpm start` for production or `pnpm dev` for development in the project directory. You should now be able to access the site at http://localhost:32464/
+5. In the `dbUrl` field, input your MongoDB URL. In the `dbName` field, input the name of the database you would like to use.
+6. (optional) If you are not using the default Redis port or you are using Redis remotely, you can configure the Redis URL with the `redisUrl` field in `credentials.json`. Also `discord_webhook` if you want to send errors remotely, useful in production to detect bugs.
+7. Make sure your Mongo and Redis instances are running, run `pnpm start` for production or `pnpm dev` for development in the project directory. You should now be able to access the site at http://localhost:32464/
```
metalcupcake5, line 36: imo this is more awkwardly worded than the old version, the changed text seems to be implying a list but stops after the 2nd action

metalcupcake5 (changes requested):

CONTRIBUTING.md
metalcupcake5: *just line 36, although I do feel the end of line 35 could be re-worded but that section was not changed in this pr

Shiiyu: While these changes don't really do anything of value nor leave the text in a way that wouldn't need further changes, I think it would be okay to merge this as it leaves the text in a slightly better state.

Shiiyu (approved):

PR #2189, opened 2024-02-07
Fix max level for Rainbow enchant

lunaynx:
Rainbow II and III exist, though it's basically just for fun beyond I, with II being once per profile and III being century raffle exclusive (but it generally sells for very cheap on AH for normal profiles), so not sure if y'all want to mark any lower levels as "max" or this is fine, up to y'all.

### DIFF
--- a/common/constants/enchantments.js
+++ b/common/constants/enchantments.js
@@ -68,6 +68,8 @@ export const MAX_ENCHANTS = new Set([
   "Protection VII",
   "Punch II",
   "Quantum V",
+  "Rainbow I",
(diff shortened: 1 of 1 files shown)


DuckySoLucky (approved):
lgtm, but imo we should put here all 3 levels, since higher levels aren't obtaineable anymore (gonna leave this as a decision for other devs)

Shiiyu:
Yeah, could probably put all three levels

DuckySoLucky:
@alexiayaa

PR #2202, opened 2024-03-08
fix: fixed issue with stats calculator

DespicableGoose:
Currently, the stats calculator adds the Bestiary milestone value multiplied by twenty to strength, and no value from the Bestiary to health:
Screenshot 2024-03-07 at 9 18 24 PM
Screenshot 2024-03-07 at 9 18 36 PM

This does not match the in game behavior, where no value from the Bestiary milestone is added to strength, and the Bestiary milestone value is directly added to health stat. This is the fix I have made, which can be observed in the screenshots below. The fix makes the calculator's behavior more closely match that of the game.
Screenshot 2024-03-07 at 9 18 47 PM
Screenshot 2024-03-07 at 9 18 58 PM

### DIFF
--- a/public/resources/ts/calculate-player-stats.ts
+++ b/public/resources/ts/calculate-player-stats.ts
@@ -49,7 +49,7 @@ export function getPlayerStats() {
 
   // Bestiary
   if (calculated.bestiary?.milestone !== undefined) {
-    stats.strength.bestiary = calculated.bestiary.milestone * 2;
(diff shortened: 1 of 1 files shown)


DuckySoLucky (approved):
lgtm, didn't test

ty, completely missed that when the bestiary v2 update happened

metalcupcake5 (approved):

Issue #2212, opened 2024-03-26
Missing info about the Diamond Essence shop and max Taming level


Peng1104:
**Describe the bug**
2 Bugs:
- Missing information about the Diamond Essence shop.
- Taming is showed max for 50 but the max is 60.

**Profile/Location**
Any profile page.

**Expected behavior**
Information about the Diamond Essence Shop in the profile page and taming max level set to 60

DespicableGoose:
Diamond Essence information was added in #2203.

PR #2213, opened 2024-03-26
Fix Taming skill max level

Peng1104:
Fix one of the bugs described in #2212

### DIFF
--- a/src/stats/skills.js
+++ b/src/stats/skills.js
@@ -76,8 +76,7 @@ export async function getSkills(userProfile, hypixelProfile, profileMembers) {
     carpentry: hypixelProfile.rankText
       ? constants.DEFAULT_SKILL_CAPS.runecrafting
           : constants.NON_RUNECRAFTING_LEVEL_CAP,
-    taming: constants.MAXED_SKILL_CAPS.taming,
-    //taming: constants.DEFAULT_SKILL_CAPS.taming + (hypixelProfile.achievements.? ?? 0),
+    taming: Math.max(hypixelProfile.achievements.skyblock_domesticator, 50),
   };
 
   return await getLevels(userProfile, profileMembers, hypixelProfile, levelCaps);
--- a/src/stats/skills.js
+++ b/src/stats/skills.js
@@ -12,10 +12,7 @@ async function getLevels(userProfile, profileMembers, hypixelProfile, levelCaps)
     }, 0);
 
     Object.assign(skillLevels, {
-      taming: getLevelByXp(SKILL.SKILL_TAMING, {
-        skill: "taming",
-        cap: Math.max(hypixelProfile.achievements.skyblock_domesticator, 50),
-      }),
+      taming: getLevelByXp(SKILL.SKILL_TAMING, {skill: "taming", cap: levelCaps.taming }),
       farming: getLevelByXp(SKILL.SKILL_FARMING, { skill: "farming", cap: levelCaps.farming }),
       mining: getLevelByXp(SKILL.SKILL_MINING, { skill: "mining" }),
       combat: getLevelByXp(SKILL.SKILL_COMBAT, { skill: "combat" }),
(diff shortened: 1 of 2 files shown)


src/stats/skills.js
```diff
@@ -12,10 +12,7 @@ async function getLevels(userProfile, profileMembers, hypixelProfile, levelCaps)
     }, 0);
 
     Object.assign(skillLevels, {
-      taming: getLevelByXp(SKILL.SKILL_TAMING, {
-        skill: "taming",
-        cap: Math.max(hypixelProfile.achievements.skyblock_domesticator, 50),
-      }),
+      taming: getLevelByXp(SKILL.SKILL_TAMING, {skill: "taming", cap: levelCaps.taming }),
```
metalcupcake5, line 15: raising taming cap requires submitting pets to george, so this must be determined by achievement level

metalcupcake5 (changes requested):

src/stats/skills.js
Peng1104: The problem is that this don't work, since the update where you can give the pets to George in a random order.

Peng1104: The Domesticator achievement only give information about the 5/10/15/20 and 25 taming levels, nothing to do whit George (https://wiki.hypixel.net/Achievements)

DuckySoLucky: yeah but it's still a lot better than no cap, now it won't be accurate at all, imo this PR shouldn't be merged

Peng1104: @DuckySoLucky then way is farming so? As farming also has a skill cap, you need to get gold in a farming collection and then buy the extra farming level whit a gold medal. There is no check for that you have it on the code but It dosent work, the same for taming the website says I am capped but I am not.

Peng1104: > @DuckySoLucky then way is farming so? As farming also has a skill cap, you need to get gold in a farming collection and then buy the extra farming level whit a gold medal. There is no check for that you have it on the code but It dosent work, the same for taming the website says I am capped but I am not.

never mind farming is working fine but taming is not, why not set cap 50 directly then?

DuckySoLucky: > @DuckySoLucky then way is farming so? As farming also has a skill cap, you need to get gold in a farming collection and then buy the extra farming level whit a gold medal. There is no check for that you have it on the code but It dosent work, the same for taming the website says I am capped but I am not.

~~what are you talking about? farming works?~~
nvm just saw msg below

DuckySoLucky: > > @DuckySoLucky then way is farming so? As farming also has a skill cap, you need to get gold in a farming collection and then buy the extra farming level whit a gold medal. There is no check for that you have it on the code but It dosent work, the same for taming the website says I am capped but I am not.
> 
> never mind farming is working fine but taming is not, why not set cap 50 directly then?

cuz using the achievement is better than settings it to just 50 or 60, and also we at SkyHelper do it that way as well so we should stay consistent in my opinion

DuckySoLucky: > > @DuckySoLucky then way is farming so? As farming also has a skill cap, you need to get gold in a farming collection and then buy the extra farming level whit a gold medal. There is no check for that you have it on the code but It dosent work, the same for taming the website says I am capped but I am not.
> 
> never mind farming is working fine but taming is not, why not set cap 50 directly then?

cuz using the achievement is better than settings it to just 50 or 60, and also we at SkyHelper do it that way as well so we should stay consistent in my opinion

Peng1104: Ok I will make the change back but the Max level for taming need to be 60 or else you will have a gold bar for not a absolute max level

Peng1104: Sorry for my confusion now that I had a deeper look in the code I got my mistake, I was assuming that level Cap == Max Level, adding a taming max level on the constants already fixed the error no need to change this part.

DuckySoLucky, line 15: that's gonna just remove cap completely?

DuckySoLucky, line 15: that's gonna just remove cap completely?

DuckySoLucky (changes requested):

src/stats/skills.js
Peng1104, line 15: > that's gonna just remove cap completely?

No just moved the cap for the "levelCaps" for consistency, same way farming / carpentry is done (Line 79)

DuckySoLucky (approved):
lgtm

metalcupcake5 (approved):

Shiiyu (changes requested):
Prettier has failed on src/stats/skills.js

PR #2244, opened 2024-04-22
fix: display secrets per profile and not from achievement

Shiiyu:
Username: MariMC
Profiles: Raspberry (Ironman) & Lime

Before:
image
image

After:
image
image


DuckySoLucky (approved):
lgtm, tested.

*off topic but if u don't mind can u go thru other ejs files and use the formatter u used to get rid of spacing. also can you lmk which extension that is*

PR #2254, opened 2024-05-10
feat: Add Power Relic to accessoryUpgrades

jani270:
Could not test it, but i dont think its needed in this case.


DuckySoLucky (approved):
lgtm, didn't test. 

can you add new chocolate accessories as well? you can copy constants from bug reports channel at SkyCrypt

DuckySoLucky (approved):

Issue #2201, opened 2024-03-03
Tecture Pack inconsistency


kada49:
**Describe the bug**
Rift collections have custom tp icons even with just vanilla tp enabled.

**Profile/Location**
https://sky.shiiyu.moe/stats/kada49/Raspberry#Collections

**Expected behavior**
SHow player heads like in the boss collections in the category below.

**Screenshots**
![Screenshot from 2024-03-03 20-39-47](https://github.com/SkyCryptWebsite/SkyCrypt/assets/45010661/d216fecb-c26c-4728-a806-5614ad114929)
![WhatsApp Image 2024-03-03 at 20 49 14](https://github.com/SkyCryptWebsite/SkyCrypt/assets/45010661/941c30f3-9e80-4e2a-b199-7c6a030733b0)

**Desktop (please complete the following information):**

- OS: Fedora 39 on Linux 6.7.6
- Browser: Mozilla Firefox
- Version 123.0

**Smartphone (please complete the following information):**

- Device: Samsung Galaxy A52 4G (SM-A525F)
- OS: Android 13, One UI 5.1
- Browser: Samsung Internet
- Version: 24.0.3.4

DespicableGoose:
I can confirm that this issue also occurs on desktop. The issue is also present for the Magmafish, Chili Pepper, Sulphur, Hard Stone and Gemstone collections. I will look further into this issue to see if I can solve it.

kada49:
I also found that the mithril collection has the same issue

DespicableGoose:
I have just identified the source of the issue. The issue stems from the file following file: [src/scripts/update-collections.js](https://github.com/SkyCryptWebsite/SkyCrypt/blob/development/src/scripts/update-collections.js#L22), more specifically on line 22. This code automatically takes the ID of item (from API), and adds that as the texture of the item. I'll see if I can fix this behavior at a later time, unless you or someone else is willing to resolve this issue soon.

DespicableGoose:
Turns out, I have no idea how to test changes with texture packs (when I launch development servers it has no working textures), so I can't really help with this, at least not with the knowledge I have right now.

PR #2260, opened 2024-06-18
fix: Collections had custom icons even with only vanilla enabled

DespicableGoose:
Many icons had custom texture pack icons even with only the vanilla texture pack enabled, as described in #2201. This pull request fixes those issues, as shown in the following screenshot:
Screenshot 2024-06-18 at 10 04 05 AM


DespicableGoose:
Note: This is a pretty bad way to solve the problem, I'll make a new PR with a better solution later, so don't merge this.

PR #2268, opened 2024-08-21
Update accessory constants

Ingrim4:
* Added the missing accessory upgrades.
* Changed `BROKEN_PIGGY_BANK` and `CRACKED_PIGGY_BANK` to be aliases of PIGGY_BANK, reflecting that these items are not upgrades of each other but rather the *same* accessory.


DuckySoLucky (approved):
lgtm, didn't test

ty for pr :heart:

PR #2277, opened 2024-09-25
Add cache times to credentials.json

ptlthg:
Adds configurable cache times (in seconds) to the credentials file. 

While I don't really think the credentials file is a great place for this, there's not really a better spot and this functionality definitely needed to go somewhere.

I also added a `cachetimes.js` file which exports helper functions to check if a cache on a specific resource is expired. These functions are then used to make the cache checks in `lib.js` much easier to understand.

Default cache times might want to be changed, there were some very confusing checks that I replaced with this PR, so make sure to double-check those.

Right now they're defined in the credentials file like this:
```js
cacheSeconds: {
  profiles: 60 * 5, // 5 minutes
  bingoProfile: 60 * 5, // 5 minutes
  museum: 60 * 30, // 30 minutes
  guild: 60 * 60 * 24, // 24 hours
},
```

### DIFF
--- a/src/cachetimes.js
+++ b/src/cachetimes.js
@@ -1,0 +1,47 @@
+import credentials from './credentials.js';
+
+/**
+ * Check if a resource is expired based on the cache time
+ * @param {number} unixMs Unix timestamp in milliseconds
+ * @param {number} cacheTime The cache time in seconds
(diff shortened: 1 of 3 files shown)


Shiiyu (approved):
lgtm, yet to test, will do tomorrow

PR #2289, opened 2024-10-14
fix: fastest kill time displaying incorrectly when below 1s and tooltips displaying without there being anything to display

### LINKED ISSUE #2250: Dragons killed in less than 1 second are shown without the "0." in front.
**Describe the bug**
In Misc, Dragons section, if you have helped killing a dragon and it died in less than 1 second, it is shown without the "0." front, so if you have a dragon killed in 0.950 seconds, as in my case, it will show as if it were 950 seconds.

**Profile/Location**
[Rubyrak on Watermelon](https://sky.shiiyu.moe/stats/Rubyrak/Watermelon#Misc)

**Expected behavior**
Show, at least in this case, 0.950s instead of 950s

**Screenshots**
image
image
I know that the second image does not show the decimals, but it looks like it takes less than 1 second.

- OS: [Windows 11]
- Browser [Google Chrome]
- Version [124]

averithefox:
Fixes #2250

### DIFF
--- a/views/sections/stats/misc/end_island.ejs
+++ b/views/sections/stats/misc/end_island.ejs
@@ -9,7 +9,7 @@
 
   <p class="stat-raw-values">
     <!-- Most Damage -->
-    <% if (calculated.misc.dragons.most_damage !== undefined) { %>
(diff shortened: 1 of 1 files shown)


DuckySoLucky (approved):
lgtm

PR #2290, opened 2024-10-15
Replace Plancke integration with nadeshiko

niqumu:
I propose replacing the Plancke integration on SkyCrypt with an integration to the Hypixel stats site [nadeshiko](https://nadeshiko.io). nadeshiko is a more modern, accessible, feature-rich, and efficient alternative to Plancke.

Plancke is often slow to update with new games, and is missing some features/stats entirely in some places. For example, Plancke is missing Pit entirely, a game which many SkyBlock players also enjoy. Plancke is missing many important features when compared to nadeshiko, such as questing and a good cards system. 

In testing, nadeshiko is much faster and more efficient than Plancke. nadeshiko updates almost instantly with new games and achievements, and it shares a similar design language to SkyCrypt, making it a more sensible integration.  

_Full disclosure: I am a developer of the nadeshiko project._

### DIFF
--- a/views/stats.ejs
+++ b/views/stats.ejs
@@ -466,8 +466,8 @@ const metaDescription = getMetaDescription()
 
         <guild-button class="additional-player-stat"></guild-button>
 
-        <!-- Placnke Button-->
-        <a href="https://plancke.io/hypixel/player/stats/<%= calculated.display_name %>" target="_blank" rel="noreferrer" class="additional-player-stat external-link">Plancke</a>
+        <!-- nadeshiko Button-->
+        <a href="https://nadeshiko.io/player/<%= calculated.display_name %>" target="_blank" rel="noreferrer" class="additional-player-stat external-link">nadeshiko</a>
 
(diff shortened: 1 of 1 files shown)


Shiiyu:
Thank you for your patience regarding this pull request. After discussing with the maintainers, we've decided not to move forward with adding Nadeshiko in its current state. We see a lot of potential in Nadeshiko as a replacement for Plancke, such as through the addition of quests and the pit, but Nadeshiko falls short in several places such as lacking several leaderboards, features, and stats available on Plancke; having significant issues with readability; the lack of consideration towards mobile usage; and other small things affecting the user experience.

For us to consider Nadeshiko, the website needs to include all of the features Plancke currently offers and a better user experience in addition to the improvements and new features Nadeshiko already brings. That being said, we think that Nadeshiko has the potential to become a true replacement to Plancke, and we'd be more than happy to revisit adding Nadeshiko in the future once improvements are made to the website.

PR #2314, opened 2024-11-05
Fix: new forgeable pets able to display forge timing

gylfirst:
Fix the forgeable pets to display their forge times:
- Remove the `LVL_1_LEGENDARY_` from the `constants/forge.js` file for all the pets.
- Change the display for pets in the `stats/mining.js` file.

I simplified the name in the constants, I was thinking about adding `Legendary` before the pet name, but as they are all of this rarity, I don't know if it's nice. I am waiting for a feedback on this.

EDIT: I modified the armadillo pet as well, since the abilities changed with the last mining update.
EDIT 2: I added the mythic rabbit ability.


DuckySoLucky (approved):
lgtm, didn't test

Shiiyu (changes requested):
When calculating forge times for pets, bejeweled collar has a level appended to it because it is under the pet category, despite it not being a pet. After accounting for this, I'll merge the pr.

Shiiyu (approved):

PR #2316, opened 2024-11-09
Updated Museum items

Leomaradan:
Updated list of Museum Items, with all children and armor_to_id ids


DuckySoLucky (approved):
lgtm, didn't test. probably should be rewritten to use items endpoint so there's no need for this. I might be able to do this next weekend but I'm not promising anything

PR #2327, opened 2025-01-28
Update accessory ignore/upgrade lists

lunaynx:
- No longer ignore General's Medallion as it is now obtainable
- Add talisman upgrades:
  - IQ Point -> 2 IQ Points
  - Talisman of the Century -> Ring of the Century
  - Seal Talisman -> Seal Ring
- Ignore new Rift-only accessories


DuckySoLucky (approved):
Lgtm, didn't test

